PR #15, opened 2017-03-10
Reorganize DB schema to handle other data sources (e.g. Pagefreezer)

Mr0grog:
~**DO NOT MERGE YET**~

This is work-in-progress for migrating the DB here to be a little less Versionista-specific (without losing anything we are currently tracking). It’s based on a discussion I had with @danielballan and should ultimately match his schema changes here: https://github.com/danielballan/web-monitoring-processing/compare/master...danielballan:db-alignment

(Hoping to find time to finish this later today, 3/10, if possible.)


Mr0grog:
Replaced this branch with some redone work I did while traveling this week. Since the structure of data is now significantly different, I’ve moved the API over to `/api/v1/*`. @lightandluck the old API routes still work *and* still return data in the old format (except IDs are now UUID strings instead of integers), so slowly switching over after this has merged should be doable.

New API routes (GET unless otherwise specified) are:

- `/api/v1/pages`: List tracked pages.

- `/api/v1/{page_id}`: Get info about a specific page. This is largely a subset of what we used to have for each page—some extra data about where it lives in Versionista, etc, is on the `version` object, since we could have versions from two different sources (e.g. Versionista, PageFreezer) for a given page.

- `/api/v1/pages/{page_id}/versions`: List of all versions for a given page.

- `/api/v1/pages/{page_id}/versions/{version_id}`: A specific version. The `links` object points back to the page that contains the version. There are two new fields here: `source_type` is a string identifying where the version came from (right now the only value is `versionista`). `source_metadata` is a JSON object whose data depends on the source type. For versionista, it has:

    - `account`: Versionista account name—same as we used to have in `page.versionista_account`.
    - `site_id`: ID of the site in Versionista
    - `page_id`: ID of the page in Versionista
    - `version_id`: ID of the version in Versionista
    - `page_url`: URL to the Versionista view showing all the versionis of this page
    - `diff_with_previous_url`: URL to diff view in Versionista
    - `diff_with_first_url`: URL to diff view in Versionista
    - `diff_length`: Size of the diff in characters
    - `diff_hash`: Hash of the Versionista-generated diff (`diff_with_previous_url`)

- `/api/v1/pages/{page_id}/versions/{version_id}/annotations`: Shortcut to `/api/v1/pages/{page_id}/versions/{version_id}/changes/{change_id}/annotations` for the change between this version and the previous version.

- `POST /api/v1/pages/{page_id}/versions/{version_id}/annotations`: Shortcut to `/api/v1/pages/{page_id}/versions/{version_id}/changes/{change_id}/annotations` for the change between this version and the previous version.

- `/api/v1/pages/{page_id}/versions/{version_id}/changes`: List of changes between another version and this version. This version will always be identified the change’s `uuid_to`. `uuid_from` is always an *earlier* version. Changes exist largely to connect a set of to two versions.

  **Changes don’t exist until they’ve been annotated.** That is: `/api/v1/pages/123/versions/abc/changes` will usually be an empty list until you have POSTed an annotation to `/api/v1/pages/123/versions/abc/annotations`.

  Changes as a list mainly exist in order to hold onto annotations that might be invalidated if a new version is later added that belongs between two existing versions, thereby splitting the annotatable change in two. As we integrate PageFreezer, I hope we can find smart ways to avoid this, but it’s a possibility we are accounting for here.

- `/api/v1/pages/{page_id}/versions/{version_id}/changes/{change_id}`: Details for a particular change. Mainly, that is the `current_annotation`, a list of annotations, and a `priority` (between 0 and 1).

- `/api/v1/pages/{page_id}/versions/{version_id}/changes/{change_id}/annotations`: List of annotations for this change.

- `POST /api/v1/pages/{page_id}/versions/{version_id}/changes/{change_id}/annotations`: Add a new annotation for a change. The POST body should be a JSON object.

Would love any feedback on this before merging. (And continued feedback for refinement afterwards too, of course.)

danielballan:
FWIW, having both `version` and `change` in one request is not what I expected. I would have expected:

`POST /api/v1/pages/{page_id}/changes/{change_id}/annotations`

or, if you prefer to refer to versions directly and avoid having a Change table,

`POST /api/v1/pages/{page_id}/changes/{version_from_id}/{version_to_id}/annotations`

Mr0grog:
Hmmm, asking for all the changes pointing to a given version seemed to me the natural way to go about this. I feel like the question to be answered most of the time here is “what annotations did I make on the difference between this version and some previous version?” That is, I think the case where you know (or care) about *both* `version_from` and `version_to` is rare. So I focused the API around `version_to`.

I think the other common question is “what are all the annotations between time A and time D.” I guess I felt like it was easier/more straightforward to ask for:

- `? → B`
- `? → C`
- `? → D`

Than for:

- `A → B`
- `A → C`
- `A → D`
- `B → C`
- `B → D`
- `C → D`

That said, `GET|POST /api/v1/pages/{page_id}/changes/{version_from_id}..{version_to_id}/annotations` seems kinda nice.

Thinking about this more/typing this out is also causing me to lean vaguely towards your suggestion of removing `change` and giving annotations a `from` and `to`. I guess I worry if we do go back to having some other kind of `A → B` item, it’ll be a pain not to have a single object to tie them to. It’s likely I am worrying about a concern that won’t be a real issue, though (either this won’t happen or won’t really be a pain).

Mr0grog:
Hmmm, API-wise, maybe…

Same as before:

- `GET /api/v1/pages`
- `GET /api/v1/pages/{page_id}`
- `GET /api/v1/pages/{page_id}/versions`
- `GET /api/v1/pages/{page_id}/versions/{version_id}`

But then, differently:

- `GET|POST /api/v1/pages/{page_id}/versions/..{version_to_id}`: Things relating to `[previous] → version_to_id`
- `GET|POST /api/v1/pages/{page_id}/versions/{version_from_id}..{version_to_id}`: Things relating to `version_from_id → version_to_id`

And get rid of change as a concept (relatedly, even if we do keep `change`, it’s primary key should probably just be `uuid_from+uuid_to`, not some other UUID).

I like the `..` (or `...`) as clearer “from A to B” indicators than `/`, which are more representative of nesting or ownership.

Mr0grog:
Enh, maybe `/api/v1/pages/{page_id}/changes/[whatever]` is better than making `versions/[whatever]` do double-duty. Sorry for the stream-of-consciousness.

danielballan:
I agree with that. I like the `..` notation, and I agree that `changes` is better than making `versions` do double duty. So, the API would then be:

* `GET|POST /api/v1/pages/{page_id}/changes/..{version_to_id}: Things relating to [previous] → version_to_id`
* `GET|POST /api/v1/pages/{page_id}/changes/{version_from_id}..{version_to_id}: Things relating to version_from_id → version_to_id`

I think I'm in favor of keeping Change as a *concept* but ditching it as a table to save a database lookup.

Mr0grog:
OK, should this be:

`GET|POST /api/v1/pages/{page_id}/changes/{version_from_id}..{version_to_id}`

or

`GET|POST /api/v1/pages/{page_id}/changes/{version_from_id}..{version_to_id}/annotations`

Now realizing something else got lost in here: `priority`, which is currently on `change`. Should it move to `version`? Does it become a specialized annotation? Something else?

danielballan:
Yeah, I think that 'priority' becomes a specialized annotation. Here's an argument for that: In the future (not in the initial production deployment) we'll probably want to sort/filter the changes that the user sees by:

* what the automated processing pipelines say about the Change
* what previous human analysts (if any) have said about the Change

And it seems that all of this will roll together into how we prioritize/filter it.

danielballan:
I forgot to answer your question: I would say keep `/annotations` on there. I don't have a specific idea of any other endpoints we might want for a given Change, but it seems very possible we'll think of some later.

Also, circling back to an earlier comment, I agree that Diffs are more ephemeral than Pages/Versions/Annotations and that the SQL db is maybe not the best place to store that info. Perhaps the Python backend should make Diffs available at a REST endpoint of its own? In my mind we haven't firmly decided whether Python communicates with Rails via the SQL db, REST endpoints, or both.

Mr0grog:
> I would say keep /annotations on there.

Hmmm, should this all really just be:

- `GET|POST /api/v1/pages/{page_id}/annotations/..{version_to_id}`
- `GET|POST /api/v1/pages/{page_id}/annotations/{version_from_id}..{version_to_id}`

i.e. should we flatten this by one level?

danielballan:
I think I prefer the original way because it reflects the "has a" relationships accurately. But this way clearly works too. My taste in REST APIs is not very well developed so I would trust your instincts here.

Mr0grog:
I think the difference is in whether there is a conceptual “change” that has annotations or if there are just free-floating annotations for a page and `{version_from_id}..{version_to_id}` is merely a way of filtering them. Still a reliable “has a,” just the *what* has is different.

e.g. Github comparison URLs: https://github.com/edgi-govdata-archiving/web-monitoring-db/compare/6432cbfdfaca7f5d3ff3bc1a9b0f0b65756012f6...125d845cf8c6c059afd76d0ba1925067cd6b6f4a

Anyway, let’s just settle it and say `/change/{id}...{id}/annotations` is the way to go. Done.

danielballan:
That's pretty convincing; I'm happy either way.

Mr0grog:
OK, I think I’m going to work towards making this merge-able tonight and whatever time I can find tomorrow. Since nobody working on UI has had a chance to comment, I think I’m going to rename the new API `v0` instead of `v1` to indicate that it’s not necessarily stable and will make changes as necessary for the needs of UI and any other clients (until at least both this and the UI have shipped a stable, usable v1 for analysts).

Data is formatted similar to [JSON:API](http://jsonapi.org), with a `links` and `data` section (or an `errors` array when there is an error), but data is not nested down in an `attributes` property. (Should we change this?)

To summarize for the API, I think we’ve got:

- `/api/v0/pages`: List all tracked pages. Responds with JSON:

    ```json
    {
      "links": {
        "first": "/api/v0/pages?page=2",
        "last": "/api/v0/pages?page=10",
        "prev": "/api/v0/pages?page=1",
        "next": "/api/v0/pages?page=3"
      },
      "data": [
        // list of page objects, as in /api/v0/pages/{page_id}
      ]
    }
    ```

- `/api/v0/pages/{page_id}`: Get info about a specific page. This is largely a subset of what we used to have for each page—data about where it lives in Versionista, etc, is now on the `version` object, since we could have versions from two different sources (e.g. Versionista, PageFreezer) for a given page. Responds with JSON:

    ```json
    {
      "data": {
        "uuid": "5005fdbd-8cce-439d-8c17-536b9f245489",
        "url": "https://energy.gov/oe/information-center/library/reports-and-other-documents",
        "title": "Reports and Other Documents | Department of Energy",
        "agency": "DOE",
        "site": "DOE - energy.gov/oe",
        "created_at": "2017-03-17T21:03:56.364Z",
        "updated_at": "2017-03-17T21:03:56.364Z",
        "latest": {
          // data for the most recently captured version. Same as:
          // /api/v0/pages/{page_id}/versions/{version_id}
        }
      }
    }
    ```

- `/api/v0/pages/{page_id}/versions`: List of all versions for a given page. Responds with JSON:

    ```json
    {
      "links": {
        "first": "/api/v0/pages/5005fdbd-8cce-439d-8c17-536b9f245489/versions?page=1",
        "last": "/api/v0/pages/5005fdbd-8cce-439d-8c17-536b9f245489/versions?page=1",
        "prev": null,
        "next": null
      },
      "data": [
        {
          // data for the most recently captured version. Same as:
          // /api/v0/pages/{page_id}/versions/{version_id}
        }
      ]
    }
    ```

- `/api/v0/pages/{page_id}/versions/{version_id}`: A specific version. The `links` object points back to the page that contains the version. There are two new fields here: `source_type` is a string identifying where the version came from (right now the only value is `versionista`). `source_metadata` is a JSON object whose data depends on the source type. For versionista, it has:

    - `account`: Versionista account name—same as we used to have in `page.versionista_account`.
    - `site_id`: ID of the site in Versionista
    - `page_id`: ID of the page in Versionista
    - `version_id`: ID of the version in Versionista
    - `page_url`: URL to the Versionista view showing all the versionis of this page
    - `diff_with_previous_url`: URL to diff view in Versionista
    - `diff_with_first_url`: URL to diff view in Versionista
    - `diff_length`: Size of the diff in characters
    - `diff_hash`: Hash of the Versionista-generated diff (`diff_with_previous_url`)

  Responds with:

    ```json
    {
      "links": {
        "page": "http://web-monitoring-db.dev/api/v0/pages/6cb1dfce-9620-4b64-8545-370833050484",
        "previous": "http://web-monitoring-db.dev/api/v0/pages/6d3977b0-f3de-4e3b-a4ff-9f385b1b9e32/versions/b47830b1-3258-4afa-bf5f-a3e1815b841c"
      },
      "data": {
        "uuid": "b47830b1-3258-4afa-bf5f-a3e1815b841c",
        "page_uuid": "6cb1dfce-9620-4b64-8545-370833050484",
        "capture_time": "2017-03-05T07:57:00.000Z",
        // uri is the location of the raw content of the version that was captured on the web site
        "uri": "s3://edgi-web-monitoring-test/982d9e3eb996f559e633f4d194def3761d909f5a3b647d1a851fead67c32c9d1",
        "version_hash":"982d9e3eb996f559e633f4d194def3761d909f5a3b647d1a851fead67c32c9d1",
        "source_type": "versionista",
        "source_metadata": {
          "account": "andmbergman@gmail.com",
          "page_id": "6213490",
          "site_id": "74282",
          "page_url": "https://versionista.com/74282/6213490/",
          "diff_hash": "d143ae99dbeddeeb8ad040d3e063ed4d25aa49a980434f914673d3701a991585",
          "version_id": "51945",
          "diff_length": 796,
          "diff_with_first_url": "https://versionista.com/74282/6213490/10209074:9441075/",
          "diff_with_previous_url": "https://versionista.com/74282/6213490/10209074:10194459/"
        },
        "created_at": "2017-03-17T21:06:53.270Z",
        "updated_at": "2017-03-17T21:06:53.270Z",
        "current_annotation": {
          "test_api1": "test value3",
          "test_item": "via basic auth",
          "test_api_old": "test value"
        }
      }
    }
    ```

- `/api/v0/pages/{page_id}/versions/{version_id}/annotations`: Shortcut to `/api/v0/pages/{page_id}/changes/..{version_id}/annotations` for the annotations on the difference between this version and the previous version.

    ```json
    {
      "links": {
        "first": "/api/v0/pages/6cb1dfce-9620-4b64-8545-370833050484/changes/..b47830b1-3258-4afa-bf5f-a3e1815b841c/annotations?page=1",
        "last": "/api/v0/pages/6cb1dfce-9620-4b64-8545-370833050484/changes/..b47830b1-3258-4afa-bf5f-a3e1815b841c/annotations?page=1",
        "prev": null,
        "next": null
      },
      "data": [
        {
          "uuid": "8c83aacd-442e-402d-82f1-f89a7c91c718",
          "author_id": 1,
          "annotation": {
            "test_item": "via basic auth"
          },
          "created_at": "2017-03-17T21:06:53.422Z",
          "updated_at": "2017-03-17T21:06:53.422Z",
          "author": {
            "id": 1,
            "email": "rob+test1@robbrackett.com"
          }
        },
        // ...repeat ad nauseum
      ]
    }
    ```

- `POST /api/v0/pages/{page_id}/versions/{version_id}/annotations`: Shortcut to `/api/v0/pages/{page_id}/changes/..{version_id}/annotations` for the change between this version and the previous version.

- `/api/v0/pages/{page_id}/changes/{from_version_id}..{to_version_id}/annotations`: List of annotations on the difference between `from_version_id` and `to_version_id`. `from_version_id` can be omitted, in which case it will be considered to be the ID of the version immediately prior to `to_version_id`: `/api/v0/pages/{page_id}/changes/..{to_version_id}/annotations`

- `POST /api/v0/pages/{page_id}/changes/{from_version_id}..{to_version_id}/annotations`: Add a new annotation on the difference between `from_version_id` and `to_version_id`. The request body should be a JSON object with any annotated fields and values desired. For now, none of them will be treated specially. In the future, certain property names might get fancy special treatment, e.g. for analysis of significance or priority, for indexing, or for filtering in queries.

danielballan:
I think this is good to go for v0. The only thing I can think to add can be added later without breaking anything: eventually we will want to paginate versions too.

Mr0grog:
> eventually we will want to paginate versions too

Oh, that’s already in there. I should have edited the example to show what it would look like if there were more than one page’s worth of records (I just took the example from an actual request and nothing we have from Versionista has so many versions as to be paged yet).

Mr0grog:
Note: currently working around #24 in the Versionista import task so we don’t need to solve it in this PR, but it needs fixing before we starting bringing in data from other sources.

Mr0grog:
Debating whether I also want to remove the `Change` model here. Otherwise I think this is good to go.

Side note: now that this is a real API I can hit, I’m realizing that:

- `/api/v0/page/{page_id}/changes/{uuid}..{uuid}/annotations` is a meaningful route, but
- `/api/v0/page/{page_id}/changes/{uuid}..{uuid}/` is not
- `/api/v0/page/{page_id}/changes` is not

…and that seems not-so-great. Since we are calling this a v0, though, I‘m not going to worry about it too much for this PR.

danielballan:
Up to you. I'm fine with dropping the Change model now or later, and [your suggestion up here](https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/15#issuecomment-287903884) about annotations would fix that "no-so-great" route situation. Want to do it in this pass or a later pass?

Mr0grog:
Alright, I think I’m going to leave the change model in place and the URLs as-is because tagging is an important upcoming feature and I think we want to be tagging changes, not versions (same as annotations).

We could implement those tags as annotations (and be different from tagging pages), but bottom line, I think changes are feeling like an important concept, since they are a place to hang 1) annotations, 2) significance, and 3) tags. However we implement those, they are conceptually different ideas and, which makes changes an important concept. I’d rather keep the model around for now that flip back and forth removing and adding it.

danielballan:
Makes sense to me!

PR #16, opened 2017-03-11
AWS Move: Docker deployment

danielballan:
This PR is a proof-of-concept of deploying the app from inside Docker. The Rails app and the postgres server are run in separate containers, linked together and orchestrated by `docker-compose`. I imagine that once web-monitoring-processing is integrated we would deploy that as part of this set, potentially linked into the same postgres container.

As written, only the `deployment` RAILS_ENV is affected, which doesn't really make sense. We need to decide if we want to apply this to `production` only or to all envs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM ruby:2.4
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev nodejs
+RUN mkdir /myapp
+WORKDIR /myapp
+ADD Gemfile /myapp/Gemfile
+ADD Gemfile.lock /myapp/Gemfile.lock
+RUN bundle install
+ADD . /myapp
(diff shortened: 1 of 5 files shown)


danielballan:
Note: We discussed putting this content in the edgi-govdata-archiving/web-monitoring repo.

Mr0grog:
Turns out Heroku has Docker container support. (New to me!) I haven’t read up on the docs yet, but this might give us a smoother path for Heroku → AWS migration, too: https://devcenter.heroku.com/articles/container-registry-and-runtime

danielballan:
Oh good.

danielballan:
Superceded by https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/188 and in general the work in https://github.com/edgi-govdata-archiving/web-monitoring/issues/73, where Helm is poised to fill the role we originally guessed docker-compose might will.

PR #18, opened 2017-03-19
Readme file modification for postgres instructions

daas-ankur-shukla:
This PR holds a minor modification to the Readme file to include instructions for creating development data base and edit database.yml file to add postgres password for successful firing of rails server

This shall prevent minor difficulty faced by new comers to setup web-monitoring-db locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,13 @@ It’s a Rails app that:
    $ bundle install
    ```
 
-5. Create development database 
+5. Create development database
 
    ```sh
    $ createdb webpage-versions-db_development
    ```
 
-6. Edit the file ~\web-monitoring-db\config\database.yml. Uncomment line 35 and append your postgres database role password	
+6. Edit the file ~\web-monitoring-db\config\application.yml. replace '*****' with your login role password to connect with postgresql
 
 7. Run the following in the repo directory:
 
--- a/README.md
+++ b/README.md
@@ -25,10 +25,23 @@ It’s a Rails app that:
 
    ```sh
    $ bundle install
+   ```
+
+5. Create development database 
+
+   ```sh
+   $ createdb webpage-versions-db_development
+   ```
+
+6. Edit the file ~\web-monitoring-db\config\database.yml. Uncomment line 35 and append your postgres database role password	
+
+7. Run the following in the repo directory:
+
+   ```sh
    $ rails server
    ```
 
-5. You should now have a server running and can visit it at http://localhost:3000/
+8. You should now have a server running and can visit it at http://localhost:3000/
 
 To actually pull down new revisions in the last 24 hours from Versionista:
 
--- a/config/database.yml
+++ b/config/database.yml
@@ -24,7 +24,8 @@ default: &default
 development:
   <<: *default
   database: webpage-versions-db_development
-
+  password: postgres
+  #password: Figaro.env.LOGINROLE_PASSWORD
   # The specified database role being used to connect to postgres.
   # To create additional roles in postgres see `$ createuser --help`.
   # When left blank, postgres will use the default role. This is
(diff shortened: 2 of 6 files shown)


dcwalk:
This would address #17

README.md
```diff
@@ -25,10 +25,23 @@ It’s a Rails app that:
 
    ```sh
    $ bundle install
+   ```
+
+5. Create development database 
+
+   ```sh
+   $ createdb webpage-versions-db_development
```
Mr0grog: It would be a little better to see the config files modified to be `web-monitoring-db_development`. This name is a holdover from the previous name of this project and I should have changed it when the project was renamed :\

README.md
```diff
@@ -25,10 +25,23 @@ It’s a Rails app that:
...
+   ```
+
+6. Edit the file ~\web-monitoring-db\config\database.yml. Uncomment line 35 and append your postgres database role password
```
Mr0grog: Hmmm, we really shouldn’t suggest people modify a file managed by version control. We should add support for some other method of configuration (e.g. environment vars, a local config, etc) and recommend that.

The easiest solution (but least nice) is to just tell people to set the `DATABASE_URL` env var when running the rails server:

```sh
$ DATABASE_URL='postgresql://user:password@localhost/my_database' rails server
```

Alternatively, some good gems to help with this are [Figaro](https://github.com/laserlemon/figaro) and [dotenv](https://github.com/bkeepers/dotenv)

daas-ankur-shukla: @Mr0grog thanks for suggestions. may i work on this and suggest a revision on the PR? i would love to

Mr0grog: Great! 👍

daas-ankur-shukla: @Mr0grog i installed Figaro in the app but i am still wondering how to take the password input from user. Should we ask them to edit some other file which does not contain any sensitive info or should we take input on the prompt while installation?

Mr0grog: The idea with both Figaro and dotenv is that there is a file each person working on the app maintains that is not in version control. For Figaro, it’s `/config/application.yml` and for Dotenv, it’s `/.env`. Users would edit one of those files.

daas-ankur-shukla:
@Mr0grog my recent commits dont work. figaro env variable is not being found and a connection error persists. i comitted that code in a hurry. i'll solve it soon

config/database.yml
```diff
@@ -24,7 +24,8 @@ default: &default
 development:
   <<: *default
   database: webpage-versions-db_development
-
+  password: postgres
+  #password: Figaro.env.LOGINROLE_PASSWORD
```
Mr0grog: I think you’ve got two options here.

1. You can use ERB syntax to pull in dynamic values:

    ```yml
    password: <%= Figaro.env.LOGINROLE_PASSWORD %>
    ```
  
    …although I think a name that clarifies this is the *database* password might be an improvement over `LOGINROLE_PASSWORD`.

2. Instead of setting a `LOGINROLE_PASSWORD` environment var, set `DATABASE_URL`, which is a standard name ActiveRecord will use automatically instead of the configuration in `database.yml`:

    ```yml
    # in config/application.yml
    development:
      DATABASE_URL: 'postgres://username:password@localhost:5432/database_name'
    ```

Mr0grog: I think I would tend towards (2) above because it’s a well known, standard thing.

daas-ankur-shukla: @Mr0grog sir i have committed some changes with option 1. They are working. I went ahead with 1 as it had less changes to make. However I am reading about the comparison between 1 and 2 and shall implement the better one as final changes

Mr0grog:
@daas-ankur-shukla are you still working on this? You last message said:

> However I am reading about the comparison between 1 and 2 and shall implement the better one as final changes

…so I’m just wondering if you’re still doing that.

Mr0grog:
Closing this for now; I think we are hoping to get a nice docker setup running soon (updates to #16), so that may obviate this anyway. If somebody wants to pick this back up, though, please comment.

PR #30, opened 2017-04-25
Normalize page URLs

### LINKED ISSUE #25: Normalize page URLs to always be full, absolute URLs
@lightandluck noted that sometimes we get them imported with a scheme (e.g. `http://`) and domain, but sometimes not. This should definitely be normalized on input into the DB so that all URLs include a scheme and domain.

I *think* I fixed this in https://github.com/Mr0grog/versionista-edgi-node (it always outputs a full URL), but we still have legacy data that needs cleanup. We will also be receiving data from other sources in the future that might not have these guarantees.

- [x] Normalize URLs on model when setting/before saving
- [x] Migrate and normalize existing data

Mr0grog:
Fixes #25 and ensures that pages always have absolute URLs.

- If they are missing a scheme, it adds `https://`
- If they are missing a domain and TLD, they are marked as invalid so the page can’t be saved

This also adds a migration that fixes up existing page records.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,6 +22,12 @@ Metrics/ParameterLists:
   Exclude:
     - 'lib/versionista_service/*'
 
+Layout/AlignParameters:
+  EnforcedStyle: with_fixed_indentation
+
+Layout/IndentHash:
+  EnforcedStyle: consistent
+
 Style/EmptyLines:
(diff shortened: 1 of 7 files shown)


Mr0grog:
Rebased, since master had a huge pile of little linter fixes.

PR #32, opened 2017-05-03
WIP: Version Importing

Mr0grog:
**Work-in-Progress: Do Not Merge**

This is still kinda messy, but is functional (and running in a staging environment at https://web-monitoring-db-staging.herokuapp.com/), so I want to show work here and allow anyone to comment.

## Importing API

This adds three new endpoints:

1. `POST /api/v0/pages/{page ID}/versions` Allows you to POST a new version of the page at this URL as a JSON document.

2. `POST /api/v0/imports` Allows you to POST lots of new versions of lots of pages all at once. Imports created this way are performed on a separate worker machine managed by a job queue.

3. `GET /api/v0/imports/{import ID}` Gets the status of an import created via the above endpoint. This gives you back info like:

    ```json
    {
      "data": {
        "id": 90,
        "user_id": 1,
        "status": "complete",
        "file": "09c86c18-913c-4a31-944a-f781ba931320",
        "processing_errors": [],
        "created_at": "2017-05-02T21:00:00.591Z",
        "updated_at": "2017-05-02T21:00:11.627Z"
      }
    }
    ```

    `status` can be `pending`, `processing`, or `complete`. Poll the endpoint and check this field to await completion.
    `processing_errors` is always an array and will be filled in with any errors encountered during importing once the import is `complete`.

The POST endpoints take a JSON body formatted roughly like a version record that you get from `/api/v0/pages/{page ID}/versions/{version ID}`. Specifically, it should look like (comments removed, of course):

```javascript
{
  // ISO datetime of when the version was captured
  "capture_time": "2017-04-23T17:25:43.000Z",
  // A URI pointing to the raw content of the version. If the version is already
  // in the app's S3 bucket, it simply holds a reference to it. Otherwise, it
  // downloads the content and re-uploads it into its S3 bucket for safekeeping
  "uri": "https://edgi-versionista-archive.s3.amazonaws.com/versionista1/74304-6222353/version-10997815.html",
  // Hash of the version's raw content (not the diff)
  "version_hash": "f366e89639758cd7f75d21e5026c04fb1022853844ff471865004b3274059686",
  // Any kind of string identifying how the version was caputured
  "source_type": "versionista",
  // A JSON object with any fields you want. This gets stored as-is.
  "source_metadata": {
    "account": "versionista1",
    "site_id": "74304",
    "page_id": "6222353",
    "version_id": "10997815",
    "url": "https://versionista.com/74304/6222353/10997815/",
    "page_url": "https://versionista.com/74304/6222353/",
    "has_content": true,
    "error_code": "404",
    "diff_with_previousUrl": "https://versionista.com/74304/6222353/10997815:10987625",
    "diff_with_firstUrl": "https://versionista.com/74304/6222353/10997815:9434924",
    "diff_hash": "1b94199f9e2ac9a0b28b533a90b2126bf991a4350b473b0c35729cc4f7ade6e1",
    "diff_length": 18305
  }
}
```

For the `imports` endpoint, the POSTed data must be either a JSON array of objects like the above or a newline-delimited JSON stream (that is, JSON objects delimited by `\n` characters). The latter is a little more efficient for large uploads because the importer can parse it piecemeal.

Additionally, `import` version objects must also include four more fields:

1. `page_url`: The URL of the page this is a version of.
2. `page_title`: The title of the page this is a version of (used if a new page record needs to be created)
3. `site_agency`: The name of the government agency that manages this page (used if a new page record needs to be created)
4. `site_name`: The name of the site this page is a part of (used if a new page record needs to be created)

Note there’s not currently a way to directly upload raw version content—it has to already be on a network-accessible server somewhere. This is not ideal, but was expedient for getting this working. For Versionista data, we’ll already have it archived in an OK place, so the `web-monitoring-db` app doesn’t actually need to touch it at all. I’m open to ideas for a better protocol around uploading raw content. (zip files? multipart form encoding? base64? Some custom, non JSON binary format?)

## Architecture

An `Import` model and database table holds data about queued imports.

Single version import logic (the `POST /pages/{page id}/versions` endpiont) lives entirely in the `VersionsController`. The bigger importer’s implementation is mostly in the `ImportVersionsJob`. The two should really be unified.

The app now requires two S3 buckets (though you could configure them to be the same): one for archiving raw version content (so everything in it is public) and one for private file storage (imports are stored temporarily here for handoff to the actual import job).

The private S3 bucket doesn’t currently ever get cleaned up. That’s useful for backtracking to understand errors or for auditing, but it’ll get weighty over time; we may want to consider a periodic cleanup job.

The `Archiver` module implements the basic work of archiving the raw content for a version. Just hand it a URL and it will decide whether it needs to do anything and then do what needs to be done (e.g. download, hash, upload to S3, or any subset of that work) to archive the content at that URL.

There’s a `Archiver::Stores` module that holds an `S3` class that implements storage/retrieval logic. The idea here is that you can configure the archiver with any sort of implementation (local file store, Google cloud, Memcache, Redis). I would like to implement at least a local file storage class for testing on a dev machine (as opposed to on a server).

The private S3 bucket mentioned above _should_ be using the `Stores` module to implement its work, but is not yet.

Job queuing is managed through ActiveJob, which is using `Resque` (ActiveJob is basically a wrapper over a pluggable implementation system), which is a Redis-based queue. So the app now requires Redis.

## To Be Done

- [x] Move `Stores` module out from under `Archiver`
- [x] Add local file store for dev/test
- [x] Use a `Stores` class for private file storage when importing
- [x] More tests (pretty poor coverage right now)


Mr0grog:
Forgot to mention, when POSTing to `/api/v0/imports`, you should use a `Content-Type` that is *not* `application/json` for large imports. (I’ve been POSTing newline delimited JSON using the type `application/x-json-stream`)

Rails unfortunately auto-parses the POST body, which can slow down really big imports. I’ve been looking into it and it seems this happens at a pretty low level, so I can only turn it off if I turn it off for the whole app. I haven’t bitten the bullet and done that yet :(

danielballan:
This looks great. I hadn't run across the trick of using newline-delimited JSON chunks rather than one big array; that's smart.

> Note there’s not currently a way to directly upload raw version content—it has to already be on a network-accessible server somewhere. This is not ideal, but was expedient for getting this working.

As far as I know that's not a pressing concern. All of the sources we've been talking about (IA, PF, Sentry) can publish their raw version content at a publicly accessible URL, and a separate copy won't be necessary. Uploading raw content is a problem for several months from now, I think.

Mr0grog:
> I hadn't run across the trick of using newline-delimited JSON chunks rather than one big array

Yeah, it’s somewhat common in the Node.js community (I think because so much is built around streams), but doesn’t seem like it ever really caught on elsewhere. There’s not even really an agreed-upon mime type for it. ¯\\\_(ツ)\_/¯

>> Note there’s not currently a way to directly upload raw version content
> 
> As far as I know that's not a pressing concern. All of the sources we've been talking about (IA, PF, Sentry) can publish their raw version content at a publicly accessible URL

Sweet, that’s good to know!

Mr0grog:
OK, there’s probably a lot of time I could spend massaging this, but I think I am going to merge it now so we can deploy and get people using it.

PR #38, opened 2017-05-09
Remove old Versionista scraping code

### LINKED ISSUE #36: Remove scraper components
Versionista scraping is now handled by an external service that uses the importing API. We should get rid of the binary dependency on Phantom and all the Ruby stuff for Versionista scraping here since it is no longer used.

Mr0grog:
Versionista scraping is now handled by an external service, so we should remove it here. This also makes setup much simpler :)

Fixes #36


Mr0grog:
In the process of cleaning up the README, I realized removing the scraper left us with setup instructions that just give you an empty DB, so I also added seed data here (8 hours of both versionista accounts from today).

Mr0grog:
Note this also fixes the DB name as noted here: https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/18#discussion_r106976104, so after this PR anyone doing development will need to either rename their local database or change their local configuration to point to their current database.

PR #39, opened 2017-05-09
Upgrade Ruby (4.2.1) and Rails (5.0.2)

Mr0grog:
Would have liked to upgrade Rails to 5.1 here, but Devise is not (yet) compatible. It _should_ be released later this week, though. (https://github.com/plataformatec/devise/issues/4525)


Mr0grog:
Rebased to clean up merging with master.

Mr0grog:
Oof, looks like there is an incompatibility between Rails and the latest minitest (https://github.com/rails/rails/pull/29022). Probably best to simply hold off on this PR until Rails 5.1.1 is out.

PR #41, opened 2017-05-11
Add some querystring filters for pages and versions!

### LINKED ISSUE #13: Add filtering by site to API
It should be possible to ask only for the pages in a particular site, for example. But would probably be good to have arbitrary filtering rules.

We need site-based filtering at the very least, though, since that is how analysis work is divided up.

Mr0grog:
Pages can be filtered by:

- Site (`/pages?site={site name}`)
- Agency (`/pages?agency={agency name}`)
- URL (`/pages?url={url}` where `url` can contain `*` as a wildcard)

Versions can be filtered by:

- Hash (`/pages/{page id}/versions?hash={hash}`)
- Capture time (`/pages/{page id}/versions?capture_time={time}` where `time` is a [W3C-style ISO 8601 date/time](https://www.w3.org/TR/NOTE-datetime))

    Note `time` can also be a range, as in `time..time`, e.g. `2017-04-01T10:00Z..2017-04-01T12:00Z` for versions between 10am and noon on April 1st.

Fixes #13.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,10 +15,7 @@ AllCops:
     - vendor/**/*
 
 Metrics/ClassLength:
-  Exclude:
-    - 'lib/versionista_service/*'
-    - 'app/jobs/import_versions_job.rb'
-    - 'test/**/*.rb'
+  Enabled: false
 
 Metrics/ParameterLists:
   Exclude:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -36,6 +36,9 @@ Style/EmptyLines:
 Style/GuardClause:
   Enabled: false
 
+Style/MultilineMethodCallIndentation:
+  EnforcedStyle: indented
+
 Style/RegexpLiteral:
   AllowInnerSlashes: true
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,6 +18,7 @@ Metrics/ClassLength:
   Exclude:
     - 'lib/versionista_service/*'
     - 'app/jobs/import_versions_job.rb'
+    - 'test/**/*.rb'
 
 Metrics/ParameterLists:
   Exclude:
--- a/app/controllers/api/v0/annotations_controller.rb
+++ b/app/controllers/api/v0/annotations_controller.rb
@@ -50,7 +50,7 @@ class Api::V0::AnnotationsController < Api::V0::ApiController
       from_uuid: parent_change.from_version.id,
       to_uuid: parent_change.version.id
     })
-    api_v0_page_annotations_path(*args)
+    api_v0_page_annotations_url(*args)
   end
 
   def set_annotation
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -75,26 +75,11 @@ class Api::V0::ApiController < ApplicationController
     end
   end
 
-  def where_range_or_exact_param(collection, name, attribute = nil, &parse)
+  def where_in_range_param(collection, name, attribute = nil, &parse)
+    return collection unless params[name]
+
     attribute = name if attribute.nil?
     range = parse_unbounded_range!(params[name], name, &parse)
-    return collection unless range
-
-    if attribute.is_a?(String) && attribute.include?('.')
-      join_model = attribute.split('.')[0].to_sym
-      collection = collection.joins(join_model)
-    end
-
-    if range.is_a? Array
-      where_in_unbounded_range(collection, attribute, *range)
-    elsif range
-      collection.where("#{attribute} = ?", range)
-    end
-  end
-
-  def where_in_unbounded_range(collection, attribute, from, to)
-    collection = collection.where("#{attribute} >= ?", from) if from
-    collection = collection.where("#{attribute} <= ?", to) if to
-    collection
+    collection.where_in_unbounded_range(attribute, range)
   end
 end
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -48,4 +48,16 @@ class Api::V0::ApiController < ApplicationController
       (error.try(:has_key?, :message) && error.send(:[], :message)) ||
       error.to_s
   end
+
+  def filter_param(collection, name, attribute = nil)
+    attribute = name if attribute.nil?
+    params[name] ? collection.where(attribute => params[name]) : collection
+  end
+
+  def parse_date!(date)
+    raise 'Nope' unless date.match?(/^\d{4}-\d\d-\d\d(T\d\d\:\d\d(\:\d\d(\.\d+)?)?(Z|([+\-]\d{4})))?$/)
+    DateTime.parse date
+  rescue
+    raise Api::InputError, "Invalid date: '#{date}'"
+  end
 end
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -49,15 +49,52 @@ class Api::V0::ApiController < ApplicationController
       error.to_s
   end
 
-  def filter_param(collection, name, attribute = nil)
-    attribute = name if attribute.nil?
-    params[name] ? collection.where(attribute => params[name]) : collection
-  end
-
   def parse_date!(date)
     raise 'Nope' unless date.match?(/^\d{4}-\d\d-\d\d(T\d\d\:\d\d(\:\d\d(\.\d+)?)?(Z|([+\-]\d{4})))?$/)
     DateTime.parse date
   rescue
     raise Api::InputError, "Invalid date: '#{date}'"
+  end
+
+  def parse_unbounded_range!(string_range, param = nil)
+    return nil unless string_range
+
+    if string_range.include? '..'
+      from, to = string_range.split(/\.\.\.?/)
+      from = from.present? ? (yield from) : nil
+      to = to.present? ? (yield to) : nil
+
+      if from.nil? && to.nil?
+        name = param ? "#{param} range" : 'Range'
+        raise Api::InputError, "#{name} must have a start or end"
+      end
+
+      [from, to]
+    else
+      yield string_range
+    end
+  end
+
+  def where_range_or_exact_param(collection, name, attribute = nil, &parse)
+    attribute = name if attribute.nil?
+    range = parse_unbounded_range!(params[name], name, &parse)
+    return collection unless range
+
+    if attribute.is_a?(String) && attribute.include?('.')
+      join_model = attribute.split('.')[0].to_sym
+      collection = collection.joins(join_model)
+    end
+
+    if range.is_a? Array
+      where_in_unbounded_range(collection, attribute, *range)
+    elsif range
+      collection.where("#{attribute} = ?", range)
+    end
+  end
+
+  def where_in_unbounded_range(collection, attribute, from, to)
+    collection = collection.where("#{attribute} >= ?", from) if from
+    collection = collection.where("#{attribute} <= ?", to) if to
+    collection
   end
 end
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -26,7 +26,7 @@ class Api::V0::PagesController < Api::V0::ApiController
   def page_collection
     collection = Page.where(params.permit(:site, :agency))
 
-    collection = where_range_or_exact_param(
+    collection = where_in_range_param(
       collection,
       :capture_time,
       'versions.capture_time',
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -1,9 +1,6 @@
 class Api::V0::PagesController < Api::V0::ApiController
   def index
-    query = Page.all
-    query = query.where(site: params[:site]) if params[:site]
-    query = query.where(agency: params[:agency]) if params[:agency]
-
+    query = page_collection
     paging = pagination(query)
     pages = query.order(updated_at: :desc).limit(paging[:page_items]).offset(paging[:offset])
 
@@ -24,5 +21,21 @@ class Api::V0::PagesController < Api::V0::ApiController
 
   def paging_path_for_page(*args)
     api_v0_pages_path(*args)
+  end
+
+  def page_collection
+    collection = Page.all
+    collection = collection.where(site: params[:site]) if params[:site]
+    collection = collection.where(agency: params[:agency]) if params[:agency]
+    if params[:url]
+      query = params[:url]
+      if query.include? '*'
+        query = query.gsub('%', '\%').gsub('_', '\_').tr('*', '%')
+        collection = collection.where('url LIKE ?', query)
+      else
+        collection = collection.where(url: query)
+      end
+    end
+    collection
   end
 end
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -24,9 +24,14 @@ class Api::V0::PagesController < Api::V0::ApiController
   end
 
   def page_collection
-    collection = Page.all
-    collection = filter_param(collection, :site)
-    collection = filter_param(collection, :agency)
+    collection = Page.where(params.permit(:site, :agency))
+
+    collection = where_range_or_exact_param(
+      collection,
+      :capture_time,
+      'versions.capture_time',
+      &method(:parse_date!)
+    )
 
     if params[:url]
       query = params[:url]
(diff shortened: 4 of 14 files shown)


Mr0grog:
This is running and can be tested on [staging](https://web-monitoring-db-staging.herokuapp.com). Try it out:

[https://web-monitoring-db-staging.herokuapp.com/api/v0/pages?url=\*nps.gov/history\*](https://web-monitoring-db-staging.herokuapp.com/api/v0/pages?url=*nps.gov/history*)

danielballan:
The version time range filter is great to have. I briefly thought, "Do we want to support full regex searches, not just wildcard asterisk?" But I think asterisk is good.

I expect we will rethink 'site' and 'agency' into something a little more general in future iterations, but that's off-topic for v0.

danielballan:
One more thought: should we paginate the results sets?

```
/pages?url={url}&start=100&num=50
```

with some hard-coded max number of results that the app is willing to return in one query?

Mr0grog:
> I expect we will rethink 'site' and 'agency' into something a little more general in future iterations, but that's off-topic for v0.

Yeah, I think you're probably right on both counts.

> should we paginate the results sets?

Already works, using the `page` param:

`/pages?url={url}&page=50`

…although you can’t modify the page size right now. I should probably change that. Maybe `size` or `page_size`? Or would it be better to remove "page" from the term since it’s probably confusing in our context: `start` or `offset` and `size` or `count` or `limit`?

Mr0grog:
Noodling a little more, I'm thinking of some fairly immediate and concrete use cases this API doesn't handle well:

1. Get all versions with a particular hash (I did this for @trinberg the other day in SQL to identify all pages that were redirecting to a particular page)
2. List *all* versions from a given timeframe, not just a versions of a particular page
3. List pages that had versions in a given timeframe

I *think* all of these are needed for a basic UI that meets current analyst needs.

(1) and (2) are pretty similar—they're really about querying all versions across all pages. I'm seeing two obvious solutions here:

- `/pages/*/versions?{query}`
- `/versions?{query}` (would probably still have `/pages/{id}/versions?{query}` too)

I’m feeling the second one is more sensible, even though it expands the routes we have.

For (3), we could do any (all?) of:

- Specially support a `capture_time` query on pages
- Support all the version queries on pages (they'd return the pages that have > 0 versions matching the criteria)
- Track latest version’s capture time and hash (anything else?) directly on the page table. This might be useful metadata, though it is not as useful for querying as the previous idea. Scraping Versionista works this way and it's already a pain because if you are querying a date range with an end, it only tells you whether a page *might* have relevant versions, not whether it actually does.

Anyway, thoughts on the above? Other things that I might be missing?

danielballan:
I endorse `/versions?{query}`

Mr0grog:
Recap of discussion w/ @danielballan:

- Add `/api/v0/versions?{query}`, `/api/v0/versions/{id}`
- Don’t worry about adding page size customization for now (see #44)
- Special case a `capture_time` query on pages

lightandluck:
Is there a way to pull versions by site and created within a time frame? In prior talks you mentioned that analysts are tasked by site and they check versions every 3 days. I tried something like this: api/v0/pages?site=DOE%20-%20nrel.gov/versions?2016-02-01T10:00Z..2017-04-01T12:00Z, to get close to what analysts would want to look at, but that didn't work.

danielballan:
@lightandluck All the query string parameters have to be at the end, like so. There can only be one `?`.

```
/api/v0/versions?2016-02-01T10:00Z..2017-04-01T12:00Z&site=DOE%20-%20nrel.gov
```

That said -- after a quick test I haven't convinced myself that the `site` parameter is respected by the `/versions` controller.... checking that now.

Mr0grog:
Yeah, `site` is only on pages (sorry, was in a meeting). You can do `?capture_time={time}..{time}&site={site}` together on pages, though.

Do we want to add `site` to versions, too?

danielballan:
Do you think we'd have a larger performance penalty for querying on the join of versions + pages? There are no name collisions between the tables (other that created_at/updated_at, which I consider internal details) so there would be no ambiguity in the query string.

Mr0grog:
> Do you think we'd have a larger performance penalty for querying on the join of versions + pages?

I don’t imagine it would be larger than querying `capture_time` on page, which does effectively the same join.

danielballan:
Good point. In that case, :+1: for supporting all [public] page columns as query string parameters on `/versions`.

Mr0grog:
There is also the option to maybe have a query arg on pages like `include_versions=true`, which would give you a collection of pages, each with a collection of all matching versions:

```
GET /api/v0/pages?capture_time=2017-04-01..2017-04-02&include_versions=true
```

gets:

```javascript
{
  "links": {
    "first": "/api/v0/pages?page=1",
    "last": "/api/v0/pages?page=21",
    "prev": null,
    "next": "/api/v0/pages?page=2"
  },
  "data": [
    {
      "uuid": "af50af24-ca37-46ff-8d62-d34ee5119009",
      "url": "https://www.whitehouse.gov/briefing-room/presidential-actions",
      "title": "Presidential Actions | whitehouse.gov",
      "agency": "Whitehouse.gov",
      "site": "Whitehouse.gov - The Record - Climate",
      "created_at": "2017-05-02T23:58:29.104Z",
      "updated_at": "2017-05-02T23:58:29.104Z",
      "versions": [
        {
          "uuid": "619043cb-876d-4e88-84eb-0ae610c0114e",
          "page_uuid": "af50af24-ca37-46ff-8d62-d34ee5119009",
          "capture_time": "2017-05-05T03:35:13.000Z",
          "uri": "https://edgi-versionista-archive.s3.amazonaws.com/versionista2/74466-6239611/version-11175079.html",
          "version_hash": "0c0d0c98de6df5f2646c960187e46063ac05d476898cb85bb483bacb80d6c519",
          "source_type": "versionista",
          "source_metadata": {
          "url": "https://versionista.com/74466/6239611/11175079/",
          "account": "versionista2",
          "page_id": "6239611",
          "site_id": "74466",
          "diff_hash": "7273e42863745d1374074ca7711076112764539207d4a0d0166cc8d83ea83289",
          "version_id": "11175079",
          "diff_length": 25673,
          "has_content": true,
          "diff_with_first_url": "https://versionista.com/74466/6239611/11175079:9470552",
          "diff_with_previous_url": "https://versionista.com/74466/6239611/11175079:11157360"
          },
          "created_at": "2017-05-08T22:05:19.785Z",
          "updated_at": "2017-05-08T22:05:19.785Z"
        },
        // ...more versions
      ]
    },
    // ... more pages
  ]
}
```

Mr0grog:
> 👍 for supporting all [public] page columns as query string parameters on /versions.

As you noted earlier, the names don’t collide, but do we still want to namespace them just in case?

e.g. `/api/v0/versions?page_site=*.nasa.gov*`

Mr0grog:
Err, that namespace separator should probably not be an `_` since some attribute names have one. More like:

`/api/v0/versions?page.site=*.nasa.gov*` or `/api/v0/versions?page-site=*.nasa.gov*`

danielballan:
Yeah, namespaces are always a good idea. `-` works.

lightandluck:
Ah, thanks! `?capture_time={time}..{time}&site={site}` on pages worked for me and my use case. I don't understand the rest of the discussion, but I'm happy with what I'm able to get!

danielballan:
Cool! As far as I know, the `site` filter is not actually doing anything right now, so it might just be by luck that you are getting the results you need. In the future (when this PR is merged and deployed) you should use `page-site` (not `site`, which will do nothing) when working with the `/versions` endpoint.

danielballan:
I just noticed your comment proposing `include_versions=true`. That actually seems cleaner to me that `page-` in a version query. I guess either could be convenient in different circumstances.

Mr0grog:
Ok, ok, to sum up the last bit of convo so it’s more understandable…

We’ve added the following query arguments to three endpoints:

- `/api/v0/pages` takes:
    - `site={exact name of site}`
    - `agency={exact name of agency}`
    - `url={url pattern with * wildcards}`
    - `capture_time={time}` or `capture_time={time}..{time}` (this one is special because what is actually being tested are the pages’ _versions_, even though it is page objects that are returned)
    - *Does not include:*
        - `title`
        - `created_at`
        - `updated_at`

- `/api/v0/pages/{page_id}/versions` and `/api/v0/versions`

    These routes are exactly the same, except the former only returns versions that belong to a particular page (specified by `page_id`)

    - `hash={sha256 hash of the version’s response body}` (note the attribute of a version this corresponds to is `version_hash`, even though the name is `hash`)
    - `source_type={source_type}` (since we will soon have data from more than just versionista, this allows you to view only data from a particular source)
    - `capture_time={time}` or `capture_time={time}..{time}`
    - *Does not include:*
        - `uri` (note that is the URI of the archived content, not the URL of the page this is a version of)
        - `source_metadata` (or any property of `source_metadata`)
        - `created_at`
        - `updated_at`


## Issues

Notably, you *cannot* use the version arguments on the `/pages` endpoint or vice-versa (with the sole exception of `capture_time`). So, if you goal is to get all the versions of pages that match a URL pattern and that were captured in a particular timeframe, you can’t do it in one request; you’d need several (you *can* get all the *pages* that match that query, though).

If we wanted to enable the above scenario, there are a couple options:

1. Add all the page arguments (`site`, `agency`, `url`) as something you can use on versions, too.

2. Add a way for a query on pages to return all the matching versions of all the matching pages, e.g. by adding a `include_versions=true` argument.

    For example, if you hit the `/pages` endpoint right now, you get:

    ```javascript
    {
      "links": { /* related URLs */ },
      "data": [
        {
          "uuid": "af50af24-ca37-46ff-8d62-d34ee5119009",
          "url": "https://www.whitehouse.gov/briefing-room/presidential-actions",
          // ...other attributes of a page
          "latest": {
            // the latest version object for this page. This is the same thing
            // that you'd get if you requested:
            //   /api/v0/pages/{page_id}/versions
            // ...and took only the first result
          }
        },
        // ...more pages
      ]
    }
    ```

    …but we could make `/pages?include_versions=true` get:

    ```javascript
    {
      "links": { /* related URLs */ },
      "data": [
        {
          "uuid": "af50af24-ca37-46ff-8d62-d34ee5119009",
          "url": "https://www.whitehouse.gov/briefing-room/presidential-actions",
          // ...other attributes of a page
          "versions": [ // note `versions`, not `latest`
            // array of *all* the version objects that match this query (or just all
            // versions of the page if no query args restrict versions, like capture_time)
          ]
        },
        // ...more pages (each with more versions)
      ]
    }
    ```

3. Relatedly, would we want to make version-specific arguments (beyond just `capture_time`) also work when querying pages? So if you used a version argument (e.g. `hash` or `source_type`) when getting pages, you’d only get the pages that have versions that match.

*Finally,* if we wanted option (1) or (3), should the arguments be namespaced? That is, if you are asking for a page argument on the versions endpoint, you’d ask for `page-{argument}`, like:

`/api/v0/versions?page-url=*.nasa.gov*` to get all versions of all NASA pages.

We can do any combination of the above three options. I want to make sure we are doing whatever the UI needs, though I would like to avoid having several ways to do the same primary operation unless someone has a strong argument for doing so.

Does that make sense, @lightandluck?


## Other stuff

> As far as I know, the site filter is not actually doing anything right now, so it might just be by luck that you are getting the results you need.

FWIW, this branch is running on staging, so if that's where @lightandluck is trying it, it *should* be working. I’m _attempting_ to make staging a merge of all PR branches that are currently being worked on and are green.

lightandluck:
Sorry for the late reply, I've been on the road the last couple hours.

Unfortunately, I'm out of town without a laptop, but I will be back on
Sunday and can give this the time it deserves then.

I think I'll have to revisit the mockups that Andrew gave us, before I have
an opinion about the issues. I may be trying to replicate too closely the
current workflow and it may be unnecessary.

Everything makes sense about the additional query parameters though, if
that's what you need to merge.

Yes, I stay on staging to avoid messing with live data.

Sorry for terseness. Sent from my phone






On May 12, 2017 3:22 PM, "Rob Brackett" <notifications@github.com> wrote:

Ok, ok, to sum up the last bit of convo so it’s more understandable…

We’ve added the following query arguments to three endpoints:

   -

   /api/v0/pages takes:
   - site={exact name of site}
      - agency={exact name of agency}
      - url={url pattern with * wildcards}
      - capture_time={time} or capture_time={time}..{time} (this one is
      special because what is actually being tested are the pages’
      *versions*, even though it is page objects that are returned)
      - *Does not include:*
         - title
         - created_at
         - updated_at
      -

   /api/v0/pages/{page_id}/versions and /api/v0/versions

   These routes are exactly the same, except the former only returns
   versions that belong to a particular page (specified by page_id)
   - hash={sha256 hash of the version’s response body} (note the attribute
      of a version this corresponds to is version_hash, even though the
      name is hash)
      - source_type={source_type} (since we will soon have data from more
      than just versionista, this allows you to view only data from a
particular
      source)
      - capture_time={time} or capture_time={time}..{time}
      - *Does not include:*
         - uri (note that is the URI of the archived content, not the URL
         of the page this is a version of)
         - source_metadata (or any property of source_metadata)
         - created_at
         - updated_at

Issues

Notably, you *cannot* use the version arguments on the /pages endpoint or
vice-versa (with the sole exception of capture_time). So, if you goal is to
get all the versions of pages that match a URL pattern and that were
captured in a particular timeframe, you can’t do it in one request; you’d
need several (you *can* get all the *pages* that match that query, though).

If we wanted to enable the above scenario, there are a couple options:

   1.

   Add all the page arguments (site, agency, url) as something you can use
   on versions, too.
   2.

   Add a way for a query on pages to return all the matching versions of
   all the matching pages, e.g. by adding a include_versions=true argument.

   For example, if you hit the /pages endpoint right now, you get:

   {
     "links": { /* related URLs */ },

   "data": [
       {
         "uuid": "af50af24-ca37-46ff-8d62-d34ee5119009",
         "url": "https://www.whitehouse.gov/briefing-room/presidential-actions"
   ,
         // ...other attributes of a page
         "latest": {
           // the latest version object for this page. This is the same thing
           // that you'd get if you requested:
           //   /api/v0/pages/{page_id}/versions
           // ...and took only the first result
         }
       },
       // ...more pages
     ]
   }

   …but we could make /pages?include_versions=true get:

   {
     "links": { /* related URLs */ },

   "data": [
       {
         "uuid": "af50af24-ca37-46ff-8d62-d34ee5119009",
         "url": "https://www.whitehouse.gov/briefing-room/presidential-actions"
   ,
         // ...other attributes of a page
         "versions": [ // note `versions`, not `latest`
           // array of *all* the version objects that match this query
(or just all
           // versions of the page if no query args restrict versions,
like capture_time)
         ]
       },
       // ...more pages (each with more versions)
     ]
   }

   3.

   Relatedly, would we want to make version-specific arguments (beyond just
   capture_time) also work when querying pages? So if you used a version
   argument (e.g. hash or source_type) when getting pages, you’d only get
   the pages that have versions that match.

*Finally,* if we wanted option (1) or (3), should the arguments be
namespaced? That is, if you are asking for a page argument on the versions
endpoint, you’d ask for page-{argument}, like:

/api/v0/versions?page-url=*.nasa.gov* to get all versions of all NASA pages.

We can do any combination of the above three options. I want to make sure
we are doing whatever the UI needs, though I would like to avoid having
several ways to do the same primary operation unless someone has a strong
argument for doing so.

Does that make sense, @lightandluck <https://github.com/lightandluck>?
Other stuff

As far as I know, the site filter is not actually doing anything right now,
so it might just be by luck that you are getting the results you need.

FWIW, this branch is running on staging, so if that's where @lightandluck
<https://github.com/lightandluck> is trying it, it *should* be working. I’m
*attempting* to make staging a merge of all PR branches that are currently
being worked on and are green.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/41#issuecomment-301199103>,
or mute the thread

lightandluck:
Thanks for that summary, Rob! It really helped.

After examination of options and the needs of the UI, I think that the `include-versions=true` parameter (restricted by other queries, particularly `capture-time`) is the best option. 

The [user dashboard](https://docs.google.com/presentation/d/1zX2FtpRusbvUFESZgsaH4lkmNz7Yz8_T-65xS0tIuBU/edit#slide=id.g1b3b3a142b_0_0) specifically wants number of versions per site within a time period. Having the versions within each page record as proposed by `include-versions` option, instead of separate page records per version as it is now, would be the easiest way to tally up versions.

When displaying versions like in the list view, page information is almost always needed and queries against `/pages` like `?capture_time={time}..{time}&site={site}` already return what's necessary. At v0, the api with the `include-versions` option should allow us to complete the user dashboard and list views for the UI.

Also, as you mentioned, title is missing as a query for pages. In the mockups, it is expected to be able to search within a time frame by site, url, id, or title. We have a route for `ID` and parameters for `site` and `url`. But it doesn't look like there's anything to search by `title`. Can that be added?

Mr0grog:
> In the mockups, it is expected to be able to search within a time frame by: site, url, id, or title. We have a route for ID and parameters for site and url. As mentioned, title is missing as a query for pages. Can that be added?

First, I think it’s worth noting that the only field that really works like a *search* is `url`. `site` is exact match only. We can change that, but search-style queries are much more intense on the DB and I should really spend some time making sure I’m using the right kind of indexes and changing some other things to make sure they’re good if we really want to use them a lot.

I can add querying/searching for `title`, but will need to add another index for that to be reasonable. I am a little worried about indexing basically every column, which is pretty high overhead.

I think the question here is really: how important do we feel this is? If it’s reasonably important, I can do it and we’ll figure out how to make it not suck later. If it’s only there just because, maybe better to hold off and get something that works first to see if not having it is really a problem.

lightandluck:
Totally understand that performance is an issue. We can leave `site` as query only. As for `title`, I only asked because it was in the mockups. I'm ok with not implementing `title` until people ask for it.

Mr0grog:
Ok, so I went ahead and added:

- `include_versions=true` on pages
- `title={title}` for *exact matches* on pages
- `source_type={source_type}` on pages (realized we need this if we want to allow people to work on only Versionista or PF data or to compare how the two are working)
- `hash={hash}` on pages

Mr0grog:
Does this seem about right for merging to people?

danielballan:
Sounds like we have consensus. If anything comes up we can iterate from here.

lightandluck:
Works!

Mr0grog:
(This is now deployed to production)

Issue #34, opened 2017-05-09
Add an authorization system


Mr0grog:
Right now, we have a special `admin?` flag on users that mainly controls whether they can get into the admin area to invite or delete users. We should make the permissions system a little more granular and separately grant users:

- Admin capability
- Permission to annotate versions
- Permission to import/create versions/pages

For safety, we probably want services that import things to *only* be able to import and not create annotations, while we might want most users to *only* be able to create annotations but not import.

danielballan:
Additionally, we need to differentiate between non-admins who can annotate and who can only view. This might play into tasking too.

danielballan:
Useful gems:

* cancancan
* pundit (probably better according to @Mr0grog)

Mr0grog:
Note we now also need permission to:

- view annotations
- view at all (I think, so long as no data is just plain public yet, we want a permission for this, if for no other reason than so we can shut it off for a malicious user.)

Mr0grog:
Related question for viewing annotations—can a user who can’t view annotations see that there were annotations made on a given change? Easy answer is probably no, but worth at least clarifying here.

danielballan:
At the lowest level, we should probably make these distinctions as granular as possible and then exposed them bundled together into a small number (3-4) of roles, to make adjustments easy to make quickly.

Mr0grog:
Note also part of this is returning permissions in the API. This should probably be in `User#as_json` (which could be automatic depending on how things are implemented) or somehow added directly in the responses for `Users::SessionsController`.

We could also consider surfacing permissions in our JWT auth tokens. That might make them pretty big, though.

danielballan:
Do we actually need this for "v0"? I think we do need to address #122 (auth everywhere) but that the admin/non-admin distinction is sufficiently granular for now.

Mr0grog:
I think we do because UI now has a [really horrible hack](https://github.com/edgi-govdata-archiving/web-monitoring-ui/blob/master/src/components/change-view.jsx#L98-L105) that makes it hard to work on annotations until this is fixed. I’d have recommended @lightandluck make that part of app configuration instead if I thought we were going to defer this.

Mr0grog:
[...]

PR #48, opened 2017-05-16
Make annotations editable

### LINKED ISSUE #46: Make annotations editable
We originally wanted to produce a nice audit trail by making annotations append-only, but after discussing on a call yesterday, @danielballan and I think enabling a nicer analyst UI that doesn’t require explicit commits is probably more important than the benefits of being append-only.

The API probably shouldn’t change for this. Instead, `POST /api/v0/pages/{id}/versions/{id}/annotations` should edit an existing annotation by the current user if there is one.

Mr0grog:
When POSTing an annotation, any existing annotation by the current user on the current change gets edited/overwritten instead of appending a new annotation. The basic effect is that each change can only have one annotation per user, which is updated whenever the user posts a new annotation.

This shouldn’t really affect the API significantly; it allows a UI to push new annotations as the user works instead of waiting for them to explicitly commit (without winding up with a gazillion annotations).

Fixes #46

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -1,7 +1,7 @@
 class Page < ApplicationRecord
   include UuidPrimaryKey
 
-  has_many :versions, -> { order(created_at: :desc) }, foreign_key: 'page_uuid', inverse_of: :page
+  has_many :versions, -> { order(capture_time: :desc) }, foreign_key: 'page_uuid', inverse_of: :page
 
   before_save :normalize_url
   validate :url_must_have_domain
--- a/test/models/change_test.rb
+++ b/test/models/change_test.rb
@@ -31,4 +31,33 @@ class ChangeTest < ActiveSupport::TestCase
     change.annotate({ test_field: 'test_value' })
     assert change.persisted?, 'The change was not persisted'
   end
+
+  test 'subsequent annotations by the same user should replace the existing annotation' do
+    change = versions(:page2_v2).change_from_previous
+
+    change.annotate({ one: 'a', two: 'b', three: 'c' }, users(:alice))
+    change.annotate({ one: 'new!', three: nil }, users(:alice))
+
+    assert_equal 1, change.annotations.count, 'Multiple annotations were made'
+    # the second annotation replaces the first instead of combining with it
+    assert_equal({ 'one' => 'new!' }, change.current_annotation, 'Current annotation was not replaced')
+  end
+
+  test 'subsequent annotations with different users should create new annotations' do
+    change = versions(:page2_v2).change_from_previous
+
+    change.annotate({ one: 'a', two: 'b', three: 'c' }, users(:alice))
+    change.annotate({ one: 'new!', three: nil }, users(:admin_user))
+
+    assert_equal 2, change.annotations.count, 'The wrong number of annotations were made'
+  end
+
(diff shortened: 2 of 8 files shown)


app/models/page.rb
```diff
@@ -1,7 +1,7 @@
 class Page < ApplicationRecord
   include UuidPrimaryKey
 
-  has_many :versions, -> { order(created_at: :desc) }, foreign_key: 'page_uuid', inverse_of: :page
+  has_many :versions, -> { order(capture_time: :desc) }, foreign_key: 'page_uuid', inverse_of: :page
```
danielballan: 👍

test/models/change_test.rb
```diff
@@ -31,4 +31,33 @@ class ChangeTest < ActiveSupport::TestCase
...
+  test 'subsequent annotations with different users should create new annotations' do
+    change = versions(:page2_v2).change_from_previous
+
+    change.annotate({ one: 'a', two: 'b', three: 'c' }, users(:alice))
+    change.annotate({ one: 'new!', three: nil }, users(:admin_user))
+
+    assert_equal 2, change.annotations.count, 'The wrong number of annotations were made'
+  end
+
+  test 'subsequent annotations with no user should create new annotations' do
```
danielballan: I gather this is anticipating a scenario where annotations are submitted by some service that is not a user. Is that what you have in mind? Either way this seems like correct behavior.

Mr0grog: Enh, this is more guarding the fact that the *REST API* requires a logged in user, but under the hood, the system does not. We could change that (maybe we should change that). Don't really know why I originally wrote it that way.

I think we would still want other services to have accounts. We always want to know who/what made an annotation, I think.

danielballan: Agreed all around. We may or may not decide to represent these other services to look like "users" or like some other kind of thing. I think it depends on whether they need special access. I suspect that will not need special access and that we can treat them as users.

Mr0grog: So… make `user` a required field in the DB and model? If we need specialized access, I think I’d favor addressing that with permissions/roles as in #34.

danielballan: Yeah, that seems like the way to go.

Mr0grog: Alrighty, will get on this later tonight if I have time or tomorrow when I am back for day work.

Mr0grog: Done.

danielballan:
Good to merge.

Mr0grog:
Ha, just realized as I went to hit the merge button that I should add a DB migration to make the `author_id` column non-null. Will merge as soon as I've done that.

Mr0grog:
> The white space changes come from the Rails version bump?

I can’t imagine any other reason. My [admittedly minimal] past experience has taught me that schema files jump around quite a bit between versions. (I presume if we switch to SQL structure dumps, which we have to if we use more fancy Postgres index types for searching, would be more stable, actually.)

Mr0grog:
It also removed all the `using: :btree` and `id: :serial` arguments, so I’m assuming those must have become implicit defaults.

Mr0grog:
Yep, looks like those changes are expected: https://github.com/rails/rails/blob/v5.1.1/activerecord/CHANGELOG.md

PR #49, opened 2017-05-16
Allow a variety of booleans for include_versions

Mr0grog:
It can be:

- valueless (`?include_versions`, but not `?include_versions=`)
- t
- 1

As a bonus, this is fairly generic so we can use it for other things in the future.

### DIFF
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -49,6 +49,12 @@ class Api::V0::ApiController < ApplicationController
       error.to_s
   end
 
+  def boolean_param(param, presence_implies_true = true)
+    value = params[param]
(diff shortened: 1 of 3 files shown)


danielballan:
As long as we're adding synonyms can we throw in 'true'?

danielballan:
Oh wait I just realized that true is the basic value and these are variants. It's been a long day.

Mr0grog:
Heh, no worries. I am also fine locking this down a bit more to just valueless or `true`. I mostly wanted to add valueless support here, but figured this was a good “be flexible in what you accept” opportunity.

Mr0grog:
Also worth noting: it’s case insensitive, you can `?include_versions=True`. Hopefully this makes it so any system serializing a dictionary/hash into a querystring will get it right enough to work.

danielballan:
Good to merge?

PR #52, opened 2017-05-19
Optimize the basic page listing call in the API

Mr0grog:
While working on https://github.com/edgi-govdata-archiving/web-monitoring-ui/pull/48, I realized I was pushing the staging server pretty hard. I knew the queries being done at `/api/v0/pages` were sub-optimal, but hadn’t bothered to fix them. Now that the logger is sending me a big pile of e-mails, it seemed like a good idea to fix it :P

Anyway, the basic idea here is to turn `Page#latest` into an actual association (albeit a little bit of an oddball one) so that ActiveRecord can actually preload it. Now we perform 2 queries for the whole page instead of 101 queries.

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -2,11 +2,18 @@ class Page < ApplicationRecord
   include UuidPrimaryKey
 
   has_many :versions, -> { order(capture_time: :desc) }, foreign_key: 'page_uuid', inverse_of: :page
-  has_one :latest, (lambda do
-    relation = order('versions.page_uuid')
-    relation.select_values = ['DISTINCT ON (versions.page_uuid) versions.*']
-    relation.order('versions.capture_time DESC')
-  end), foreign_key: 'page_uuid', class_name: 'Version'
+  has_one :latest,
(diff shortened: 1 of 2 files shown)


app/models/page.rb
```diff
@@ -2,6 +2,18 @@ class Page < ApplicationRecord
   include UuidPrimaryKey
 
   has_many :versions, -> { order(capture_time: :desc) }, foreign_key: 'page_uuid', inverse_of: :page
+  has_one :latest,
+    (lambda do
+      # DISTINCT ON requires the first ORDER to be the distinct column(s)
+      relation = order('versions.page_uuid')
+      # HACK: This is not public API, but I couldn't find a better way. The
```
danielballan: Reaching below the public API every now and then is a good sign that you're doing something interesting.

danielballan: Thinking on this a bit more... this is possible with a JOIN and GROUP BY but DISTINCT is significantly better optimized. :+1: for the current solution -- just thinking out loud.

Issue #50, opened 2017-05-16
Look into e-mail service providers


Mr0grog:
E-mail is currently sent through a GMail account, which was simple and expedient at first, but really very error prone, as its security mechanisms are designed for human users interacting with a UI. We should really be interfacing (still via SMTP) with a service like:

- Postmark (https://postmarkapp.com)
- Mailgun (https://www.mailgun.com)
- Sendgrid (https://sendgrid.com)
- Mandrill (add-on to Mailchimp)
- Amazon SES (https://aws.amazon.com/ses/)
- Sendinblue (https://www.sendinblue.com)
- ?

Mr0grog:
I have experience with Mandrill before it got integrated into Mailchimp and it was OK. Have heard some good things from others about Postmark, so I would probably lean that way.

Mr0grog:
Archivers is also using Postmark.

shapironick:
This is helpful. What do you think the email volume of this would be over say a year? So I can think about $$. 

Sent on the move

Mr0grog:
[Versionista Scraper](https://github.com/edgi-govdata-archiving/web-monitoring-versionista-scraper) sends roughly 22/month (2 e-mails every three days). This project sends e-mails for:

- Sign-up Invitations (optional)
- Sign-up Confirmations
- Password reminders

Right now we have < 20 analysts, so that's a one-time cost of 40 e-mails and then, say, 20/month if everyone forgets their password once. Then it's just about supporting a much wider/broader group of people to do analysis.

So for now: ~45/month
For future: ??? but probably not a lot.

Mr0grog:
Basically our needs are absurdly small at the moment ;)

Mr0grog:
That said, we don’t have a strong need to for extra analytics or reporting, so I think it’s reasonable to view these services as a commodity if pricing is a concern.

Mr0grog:
@shapironick Also, does EDGI have organizational e-mail at `envirodatagov.org` (or somewhere else)? If so, we should really have these applications sending from e-mail addresses at that domain.

dcwalk:
Just to update on this on issue (from convo in `#tools_talk` on EDGI slack):

- EDGI currently doesn't have existing organizational email at `envirodatagov.org`. 
- We've consolidated DNS and hosting and set up team access for editing DNS records

[...]

PR #53, opened 2017-05-25
Fix totally borked invitation code generation

Mr0grog:
Well this is embarrassing.

Ridiculously, the first invitation created would wind up having no invitation code. This is because I'm an idiot and am not used to Ruby's while/if-at-the-end-of-the-line style and didn't realize the code would first be initialized to Nil instead of a generated hash value. Once we had one bad invitation in the DB, others would behave because the routine would see that their Nil code collided with the original Nil code.

Also fixes the poorly-coded fixture that masked this issue in tests.

This is one of two problems we have with sending invitations and confirming new users in production. The other is #50.

### DIFF
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -56,8 +56,10 @@ class Invitation < ApplicationRecord
   private
 
   def assign_code
-    token = generate_code while Invitation.where(:code => token).exists?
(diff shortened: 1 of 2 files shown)


danielballan:
Ha. Subtle.

PR #54, opened 2017-05-26
Send e-mails with Postmark

Mr0grog:
**DO NOT MERGE:** Postmark still has to approve the account before we can send to most public e-mail addresses. This should happen by 2017-05-26.

Addresses #50.

Setup e-mail sending via Postmark. I’ve set up an account and this is working for me locally. When deployed, this will be sending from an account at my personal domain. We’ll want to change that (see discussion about domains and e-mail etc. in the EDGI Slack #tools_talk), but user accounts are kinda broken right now and I want to at least get them working so other people can test against the live DB without me manually setting up accounts in the console or cajoling GMail into behaving.


Mr0grog:
The account is approved!

PR #60, opened 2017-05-31
Add API for getting diffs

### LINKED ISSUE #8: Add API pass-through endpoint for diffs
We probably want to eventually have a more nuanced way to calculate diffs (see also: “scraping notes” on https://github.com/edgi-govdata-archiving/versionista-outputter/issues/4#issuecomment-281800291). This should probably be a separate service and should be configurable pluggable.

See also https://github.com/edgi-govdata-archiving/differ

Mr0grog:
Fixes #8.

This is currently implemented on top of @b5’s simple diff service: https://github.com/edgi-govdata-archiving/go-calc-diff
…but should theoretically work with any HTTP differ that accepts requests with two query arguments:

- `a`: A URL to the content of the starting version
- `b`: A URL to the content of the ending version

e.g: `http://diff-it.com?a=http%3A%2F%2Fexample.com%2Fv1&b=http%3A%2F%2Fexample.com%2Fv2`

Diffs are retrieved by requesting the URL:

```
/api/v0/pages/{page_id}/changes/{version_a_id}..{version_b_id}/diff/{diff_type}?{query_params}
```

(Meant to mirror the `annotations` API)

`query_params` are simply passed along to the underlying diff service. For example, `go-calc-diff` can take a `format` argument that determines how the response is formatted (JSON, HTML document, HTML snippet):

```
/api/v0/pages/xxx/changes/yyy..zzz/diff/source?format=json
```

That returns a response like:

```js
{
  "data": {
    "page_id": "3f28a7c3-65f2-481e-9722-efccc1f38462",
    "from_version_id": "51896700-93d7-42df-8a3c-aa0282e1c8e8",
    "to_version_id": "16d82acd-01ef-4e39-a5d2-ab11daa6540e",
    "diff_service": "DIFF SERVICE NAME",
    "diff_service_version": "?",
    "content": "RESPONSE FROM DIFF SERVICE HERE"
  }
}
```

If the response was JSON data, `content` will actually an object. Otherwise, it will be a string.

We only support one `diff_type` right now (`source`), but the system is designed to be expandable. Under the hood, there is a registry that maps diff type names to actual diff service implementations. Implementations only need to respond to one method:

```rb
def diff (change_model, options = {})
  # return a string or hash that will be used in the `content` field
end
```

For now, we just have one `SimpleDiff` class, which is instantiated with a URL. It can be used to interact with any diff service that matches the same protocol as `go-calc-diff`.


Mr0grog:
Hoping to get this worked out before Thursday so Global Sprinters have a working example of diffing.

Mr0grog:
Running on staging for testing, e.g: https://web-monitoring-db-staging.herokuapp.com/api/v0/pages/76d2e99c-c1a5-44e5-9781-6622573b9444/changes/..5ef676d4-0811-4111-aa80-19d2999ab178/diff/source or https://web-monitoring-db-staging.herokuapp.com/api/v0/pages/76d2e99c-c1a5-44e5-9781-6622573b9444/changes/..5ef676d4-0811-4111-aa80-19d2999ab178/diff/source?format=json (for fancy JSON data)

Mr0grog:
Side note: this includes a wonky patch for an encoding issue in HTTParty (https://github.com/jnunemaker/httparty/issues/542), which was also causing import issues (if raw version content had to be parsed, which is not the case for most imports).

config/initializers/differ.rb
```diff
@@ -0,0 +1,6 @@
+require_dependency 'differ/differ'
+require_dependency 'differ/simple_diff'
+
+if ENV['DIFFER_SOURCE']
+  Differ.register(:source, Differ::SimpleDiff.new(ENV['DIFFER_SOURCE']))
```
danielballan: Would it be too clever to allow arbitrary many differs with arbitrary names to be configured through `ENV` by picking up all vars with names that match `DIFFER_*`? Or will new differs likely need something more complex that `SimpleDiff`? Offhand I'm not sure what they would need so the approach I'm floating seems _doable_ -- but maybe less than _clean_.

Mr0grog: > allow arbitrary many differs with arbitrary names to be configured through ENV by picking up all vars with names that match `DIFFER_*`?

Ha! I thought about the same idea and then said “nah, let’s not get ahead of ourselves.” I think I would at least change the var to `DIFFER_SIMPLE_*` if we did that.

I can imagine some differs being more complex (or maybe `SimpleDiff` will just grow)—we can‘t meaningfully diff PDFs with this, for example, but we really should eventually have that capability. There’s a lot of PDF content that is “tracked” but has no real tooling to support analyzing it. Versionista is also hitting some audio files and potentially things like CSVs. I haven't really done a full audit of content types. By nature, I’m sure PF is grabbing who-knows-what variety of content types.

Mr0grog:
Merging so we’ve got something for Global Sprint. Once we’ve got more diff services, we can come back around and consider iterating through `DIFFER_SIMPLE_*` to auto-create differ instances.

Issue #67, opened 2017-06-01
Add banner to top of home page based deployment environment


Mr0grog:
It’d be helpful to display some sort of banner on the top of pages on the test/staging site (and maybe in dev, too?) to remind someone that they aren’t looking at the production instance of the app.

For the API, we could potentially include a special key in the JSON object or an `x-environment` HTTP header (less obvious, but probably a more wise approach).

rdalin82:
I am new to the project and would like to try my hand at this issue.  Are you looking for something like the below screenshot? 

![webmonitor-env](https://cloud.githubusercontent.com/assets/5727364/26757487/648ef496-4882-11e7-9557-3ad5d3e08d6e.png)

Mr0grog:
@rdalin82 I was thinking:

- Above existing header, so it might be easier to share the same design with `web-monitoring-ui` (https://github.com/edgi-govdata-archiving/web-monitoring-ui/issues/67), and
- Maybe more colorful so it is somewhat eye-catching?

Otherwise, yeah! It would be lovely if you wanted to work on this :)

Mr0grog:
Fixed in #72.

PR #72, opened 2017-06-04
Environment banner

rdalin82:
fixes, [#67](https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/67), I haven't setup a test yet for the environment banner, if you need a test setup for it I'd recommend doing an integration test with capybara which I could setup.  I didn't want to add the dependency to your project without your input first since it would be bringing in a library just to test one thing.   I also added some color to the banner and made it center, hopefully it sticks out more.  

I added a test for the X-Environment header, its not very expansive, but I think with a little guidance I could add more tests. 

environmentbannerwithcolor

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,4 @@
 class ApplicationController < ActionController::Base
-  # pass
   before_action :set_environment
 
   private
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,3 +1,9 @@
 class ApplicationController < ActionController::Base
   # pass
+  before_action :set_environment
+
+  private
+  def set_environment
+    @env = Rails.env
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,7 @@ class ApplicationController < ActionController::Base
   before_action :set_environment
 
   private
+
   def set_environment
     @env = Rails.env
   end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -12,13 +12,11 @@
 
   <body>
     <% unless @env == "production" %>
-      <header class="page-header">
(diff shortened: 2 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,3 +1,10 @@
 class ApplicationController < ActionController::Base
   # pass
+  before_action :set_environment
```
Mr0grog: Since there’s actual code in this class now, you should get rid of the `# pass` comment :)

app/views/layouts/application.html.erb
```diff
@@ -11,10 +11,18 @@
   </head>
 
   <body>
+    <% unless @env == "production" %>
+      <header class="page-header">
```
Mr0grog: This should really be inside the pre-existing `header` element (you’re only meant to have one per HTML “section” — this is, `article`, `aside`, `nav`, `section` elements). Otherwise we should make it a plain old `div`.

test/controllers/api/v0/pages_controller_test.rb
```diff
@@ -164,4 +164,8 @@ class Api::V0::PagesControllerTest < ActionDispatch::IntegrationTest
     results = body['data']
     assert_not results.any? {|p| p.key? 'versions'}, 'Some pages have a "versions" property'
   end
+  test 'includes environment in header of response' do
```
Mr0grog: Would you add a blank line between tests, please?

Mr0grog:
Looks awesome, thanks!

rdalin82:
Really glad I could help.  I checked out the goals of the overall project and you folks are doing some terrific work. Keep up the good work!

Mr0grog:
Thanks! I just realized we don’t have a “Contributors” section in the README, so I’m about to add one. Would you like to be listed there?

rdalin82:
That would be great thanks!

Mr0grog:
Feel free to update #75 if you’d like your link to point somewhere else.

PR #79, opened 2017-06-09
ReadMe file lacks instructions for postgres - second take

### LINKED ISSUE #17: ReadMe file lacks instructions for postgres
The readme file does not have instructions to create development database and to edit the corresponding file to append their passwords for proper firing of rails server

KrzysztofMadejski:
Fixes #17
Ping @Mr0grog

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@ It’s a Rails app that:
 ## Installation
 
 1. Ensure you have Ruby 2.4.1+
+
+   You can use [rbenv](https://github.com/rbenv/rbenv) to manage multiple Ruby versions
+
 2. Ensure you have PostgreSQL 9.5+
 3. Ensure you have [Redis](https://redis.io)
 
@@ -48,7 +51,9 @@ It’s a Rails app that:
 
     That will create a database, set up all the tables, create an admin user, and add some sample data. Make note of the admin user e-mail and password that are shown; you’ll need them to log in and create more users, import more data, or make annotations.
 
-    If you’d like to do the setup manually, see [advanced setup](#advanced-setup) below.
+    If you’d like to do the setup manually see [manual postgres setup](#manual-postgres-setup) below.
+    
+    If you're getting error such as `FATAL: role "user" doesn't exist. Couldn't create database.` check [troubleshooting](#troubleshooting) below. 
 
 8. Start the server!
 
@@ -74,7 +79,7 @@ It’s a Rails app that:
    $ QUEUE=* VERBOSE=1 bundle exec rake environment resque:work
    ```
 
-## Advanced Setup
+## Manual Postgres Setup
 
 If you don’t want to populate your DB with seed data, want to manage creation of the database yourself, or otherwise manually do database setup, run any of the following commands as desired instead of `rake db:setup`:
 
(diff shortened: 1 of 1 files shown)


KrzysztofMadejski:
Hm, maybe having `DATABASE_URL` as default variable should point to [.env.example file](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/.env.example#L5)? 

The install guide don't mention copying and modifying .env.example. Should it be added in a new advanced section?

KrzysztofMadejski:
Pin @edgi-govdata-archiving/lieutenants

patcon:
LGTM!

I love the idea of at least hat-tipping to advanced stuff, so that, even if the help isn't necessarily comprehensive, new folks don't feel they're going offroading when they hit a slight snag (perhaps for the first time using postgres or something) :)

danielballan:
Looks good. Will let Rob merge when he gets a look.

README.md
```diff
@@ -95,6 +100,47 @@ User.create(
 )
 ```
 
+## Troubleshooting
+
+If you are getting errors such as `FATAL: role "user" doesn't exist. Couldn't create database.` while running `rake db:setup` or `rake db:create` then it may mean that your database is password protected. There are two ways to setup required databases:
+  
+1. (Recommended) Loosen local Postgres database security to allow local users without password
```
Mr0grog: Ok, I may have been a little unclear in other discussions. Sorry for that. I really don’t think we should ever recommend people reduce the security of their existing setup. (I don’t think security is important for a dev DB here, but if someone has any other databases in Postgres that might actually need to be secured, we’re recommending they open up a big vulnerability. That’s bad.)

Can we make this option 2 and make the other (manually create users and DBs) the recommended option?

README.md
```diff
@@ -95,6 +100,47 @@ User.create(
...
+    sudo -u postgres createuser `whoami` -ds
+    ```
+    
+    Now `bundle exec rake db:setup` command should work.
+    
+2. Create users and databases manually
+
+    ```sh
+    sudo -u postgres psql -c "CREATE USER wmdb_dev WITH PASSWORD 'wmdb_dev';"
+    sudo -u postgres createdb -O wmdb_dev wmdb_dev -E utf-8
```
Mr0grog: I totally missed this in the earlier review, but can you please use `web-monitoring-db_development` for the DB name? I’d like to make sure the DB someone winds up with by following this matches as closely as possible to the one they’d wind up with if they did the normal steps (`rake db:setup`).

README.md
```diff
@@ -95,6 +100,47 @@ User.create(
...
+    sudo -u postgres psql -c 'CREATE EXTENSION IF NOT EXISTS "uuid-ossp";' wmdb_dev
+    sudo -u postgres psql -c 'CREATE EXTENSION IF NOT EXISTS "plpgsql";' wmdb_dev
```
Mr0grog: Same here on DB naming.

README.md
```diff
@@ -95,6 +100,47 @@ User.create(
...
+    ```
+    
+    and then set `DATABASE_URL` environment variable to point to the development database:
+     
+    ```sh
+    export DATABASE_URL=postgres://wmdb_dev:wmdb_dev@localhost/wmdb_dev
```
Mr0grog: DB naming again.

KrzysztofMadejski:
@Mr0grog done!

Mr0grog:
Beautiful. Thanks so much for working on this, @KrzysztofMadejski!

Issue #85, opened 2017-06-15
Receiving duplicate records from filter query


lightandluck:
Querying the `/pages` route using `site` and `capture_time` parameters returns duplicate entries in the response.

## Steps to reproduce
1. [Sample query](https://web-monitoring-db-staging.herokuapp.com/api/v0/pages?site=State%20-%20state.gov&capture_time=2017-06-01..)
2. Compare 2nd and 3rd entries (they begin with UUID="31aba984-7faf-4ace-9c6c-1056b51b9b6a"). They are exactly the same.

## Context
In [staging UI](https://web-monitoring-gapi.herokuapp.com/loggedIn/eric@gmail.com), filtering the list by user, React will throw numerous warning messages:

```Warning: flattenChildren(...): Encountered two children with the same key, `31aba984-7faf-4ace-9c6c-1056b51b9b6a`. Child keys must be unique; when two children share a...```

React will display the first record and toss all duplicates, which isn't a blocking issue, but doesn't seem desirable. Not sure if the duplication is expected behavior from querying the DB in this way.

lightandluck:
Resolved by #86

Mr0grog:
Whoa, don’t close this until we’ve actually merged the fix!

lightandluck:
Whoops, sorry about that! Thought I saw you self-merge this PR because it seemed a minor change. Good to know it automatically closes. Thanks!

PR #86, opened 2017-06-15
Return unique records in `where_in_unbounded_range`

Mr0grog:
This is a fix for #85.

ActiveRecord does not actually try to figure out that a `joins(:some_association)` call may return multiple instances of a given record and coalesce them into a single result, so we also need to tell it to only get distinct results when auto-joining in `ApplicationRecord#where_in_unbounded_range`.

In #85, `where_in_unbounded_range` is used on pages’ association with versions to query on `capture_time`. This caused an INNER JOIN to versions, which meant we’d see duplicate page entries for every version that matched. We really only want *one* page entry if any of that pages’ versions that match.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -6,7 +6,8 @@ class ApplicationRecord < ActiveRecord::Base
 
     if attribute.is_a?(String) && attribute.include?('.')
       join_model = attribute.split('.')[0].to_sym
-      result = result.joins(join_model)
+      # Use distinct to indicate that we only want one record per joined match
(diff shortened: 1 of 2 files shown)


Mr0grog:
See also https://github.com/edgi-govdata-archiving/web-monitoring-ui/pull/79

Mr0grog:
@lightandluck You should be able to see this fixed on staging now.

lightandluck:
Thanks, looks good and is working!

Mr0grog:
Merging away, since nobody is available to review.

PR #88, opened 2017-06-20
Send hash of versions to simplediff differs

### LINKED ISSUE #84: Include version_hash (content hash) in diff queries.
The Rails server currently queries separate servers ("diffing services") with two URLs the the service should fetch and analyze. That query should include the hash of the URL's content, which is already stored in the db, so that the diffing service can verify that the content it fetches is what is expected.

Recently discussed on a quick call with @Mr0grog -- put here for tracking.

Mr0grog:
Send the version hash for each version being diffed to the differ. This allows differs to do smarter caching. Fixes #84.

### DIFF
--- a/lib/differ/simple_diff.rb
+++ b/lib/differ/simple_diff.rb
@@ -8,7 +8,9 @@ module Differ
       options ||= {}
       query = options.merge(
         a: change.from_version.uri,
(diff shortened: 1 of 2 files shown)


shapironick:
Rad! Thanks rob!

Sent on the move

PR #93, opened 2017-07-06
Add differs from web-monitoring-processing#59

danielballan:
I left DIFF_SOURCE in place in case we want to continue to run go-calc-diff alongside web-monitoring-processing similar utility for awhile.


Mr0grog:
This looks great, but can you add the env vars with descriptions to the `.env.example` file?

Mr0grog:
(Now realizing I failed to do that originally with `DIFFER_SOURCE`, shame on me)

PR #96, opened 2017-07-08
Add remote login via API

Mr0grog:
This adds the ability for remote clients to log users in and get session tokens to use with the API. It’s needed to support https://github.com/edgi-govdata-archiving/web-monitoring-ui/issues/43

- As a security measure, cookie-based sessions are unavailable if the request has a `Referer` header that does not match DB’s current domain (so authentication data must be provided with every remote request). This prevents a malicious web site from making authenticated requests on behalf of the user.

  Cookies still work for XHRs made from the same domain as DB and if there is no `Referer` header present (so the cookie jar in your non-browser script will work fine). This will not clear existing cookies; it just makes them unavailable to the app during remote requests.

- Authentication can now be done as a JSON API.

  - `POST /users/sign_in` This is our existing sign-in route. However, if you use the `Accept: application/json` header, you get back a JSON object like:

    ```json
    {
      "user": {
        "id": 2,
        "email": "user.email@example.com"
      },
      "token": "[JWT auth token]"
    }
    ```

    The included token can be used to authorize subsequent requests by specifying the header: `Authorization: Bearer [JWT auth token]` (basically the same way you use an OAuth token), so a client can keep this token instead of credentials.

  - `GET /users/session` Validates a session token or other authorization credentials (e.g. HTTP basic auth) provided in the `Authorization:` header. The response format is the same as `/users/sign_in`, but it does not include a new `token`.

Tokens are signed using an RSA key defined in the `TOKEN_PRIVATE_KEY` environment variable (should be a PEM-formatted string). We don’t currently expose the public key this generates anywhere, but we could make it available at `GET /.well-known/token_key` or something.

In the future, it might make sense to move to OAuth or something similar, but since we only have one client, I wanted to A) avoid excess complexity and B) avoid a potentially confusing scenario where users logging in to the analysis tool get forwarded to another site when logging in. Open to pushback on that, though ¯\\\_(ツ)\_/¯


Mr0grog:
Note the other place where this falls down is in revoking tokens. I think it might be OK to handle that separately. Some thoughts:

- Keep a list of valid tokens, nice and simple, OAuth-style. Logging out can remove the token from the list.
- Keep a log out date for the user, include the standard JWT `iss` (issued at) field, and reject tokens issued before the last log out date.
- (Kinda crazy) Generate an RSA keypair for every user; sign the key with that. Generate a new keypair when logging out.

danielballan:
:+1: for starting with (2) and moving to (1) in the future if we have a reason to. Is this good to merge?

Mr0grog:
Yep, this is good to go if you don't have any concerns!

danielballan:
Cool. I read through it and couldn't see any red flags.

PR #103, opened 2017-07-13
Add better (and bigger) seed data

### LINKED ISSUE #100: Add seed data with more multi-versioned pages
Our seed data only has two pages (of 858) with multiple versions:

- `https://www.blm.gov/programs/fish-and-wildlife/wildlife/about/idaho`
- `http://www.globalchange.gov/browse/reports/sap-34-abrupt-climate-change-brochure`

It should really be more realistic and be *mostly* comprised of pages with multiple versions. The current setup is making testing hard for people. (Though it did reveal https://github.com/edgi-govdata-archiving/web-monitoring-ui/issues/76)

Mr0grog:
This data is based on the first few pages of data from the production database on 2017-07-13.

- 128 Pages (15 w/ 1 version, 113 w/ multiple)
- 1000 Versions

Note these seeds also include Version UUIDs, so we can talk about seed data in common between people's installs.

Fixes #100.


danielballan:
This raises a string encoded error for me when I run it locally on OSX:

```
$ bundle exec rake db:seed
rake aborted!
Encoding::UndefinedConversionError: "\xE2" from ASCII-8BIT to UTF-8
/Users/dallan/Documents/Repos/web-monitoring-db/lib/file_storage/local_file.rb:26:in `write'
<snip>
```

It looks like, if we end up with raw binary content, we should write as it a binary file, right?

Mr0grog:
…there shouldn’t be any raw binary content here. Hmmm. Let me look into this.

Mr0grog:
Oh, I’m betting you don’t have `$ALLOWED_ARCHIVE_HOSTS` set, do you, @danielballan? (Not that you should need to, but after unsetting that I am seeing and starting to understand the issue)

Mr0grog:
@danielballan the above should fix your issue. You may also want to consider setting:

```sh
export ALLOWED_ARCHIVE_HOSTS='https://edgi-versionista-archive.s3.amazonaws.com/ https://edgi-versionista-archive.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-versionista-archive/'
```

in your environment (then your seed data will reference URLs in S3 rather than downloading and mirroring them locally).

danielballan:
I pulled your new commit (the walk-back commit 868c9eb) and seeding still runs successfully for me, with and without the `ALLOW...` env var set. Merge at will if you are satisfied with this.

PR #104, opened 2017-07-13
Add API endpoints for change models

### LINKED ISSUE #101: Expose changes in API
I originally tried to paper over the idea of “changes” in the API and keep things simple by just annotating versions. Over time, it’s become more and more obvious that this was a bad idea. We should expose actual change objects in the API. We even already have a reasonable route for this:

- `GET /api/v0/pages/{page_id}/changes` - list changes (ordered by `to_version`, then `from_version`)
- `GET /api/v0/pages/{page_id}/changes/{from_version_id}..{to_version_id}` - a single change

Should we should also remove `current_annotation` from `Version`? (Note it already exists on `Change` and I think we would keep it there). I’m feeling like using a version as a shortcut to the change between a version and its preceding version is causing more confusion than help, but would love feedback (would also remove the `GET|POST /api/v0/pages/{id}/versions/{id}/annotations` endpoints).

Note this might also implicitly solve #4, since you could just query for the list of changes because we only store changes when annotations have happened.

/cc @danielballan @lightandluck

Mr0grog:
Adds two routes (which one might have guessed at before but used to be 404s):

- `GET /api/v0/pages/{page_id}/changes` - list changes (ordered by `to_version`, then `from_version`)

    This also supports the query args:

    - `priority_gt` OR `priority_gte` - gets changes with a priority greater than (or equal to) the given value
    - `priority_lt` OR `priority_lte` - gets changes with a priority less than (or equal to) the given value

- `GET /api/v0/pages/{page_id}/changes/{from_version_id}..{to_version_id}` - a single change

And removes all annotation-related stuff from the version routes:

- Remove `GET /api/v0/pages/{page_id}/versions/{version_id}/annotations`
- Remove `POST /api/v0/pages/{page_id}/versions/{version_id}/annotations`
- Remove `GET /api/v0/pages/{page_id}/versions/{version_id}/annotations/{annotation_id}`
- Remove `current_annotation` from Version object

Fixes #101.

### DIFF
--- a/app/controllers/api/v0/changes_controller.rb
+++ b/app/controllers/api/v0/changes_controller.rb
@@ -42,25 +42,14 @@ class Api::V0::ChangesController < Api::V0::ApiController
 
   def changes_collection
     collection = Change
-    where_numeric_range_params(collection, :priority)
-  end
 
-  def where_numeric_range_params(collection, attribute)
-    gt_param = :"#{attribute}_gt"
-    gte_param = :"#{attribute}_gte"
-    lt_param = :"#{attribute}_lt"
-    lte_param = :"#{attribute}_lte"
-
-    if params[gt_param]
-      collection = collection.where("#{attribute} > ?", params[gt_param])
-    elsif params[gte_param]
-      collection = collection.where("#{attribute} >= ?", params[gte_param])
-    end
-
-    if params[lt_param]
-      collection = collection.where("#{attribute} < ?", params[lt_param])
-    elsif params[lte_param]
-      collection = collection.where("#{attribute} <= ?", params[lte_param])
+    if params[:priority]
+      collection =
+        if params[:priority].match?(/^\d/)
+          collection.where(priority: Float(params[:priority]))
+        else
+          collection.where_in_interval(:priority, params[:priority])
+        end
     end
 
     collection
--- a/app/controllers/api/v0/changes_controller.rb
+++ b/app/controllers/api/v0/changes_controller.rb
@@ -1,0 +1,80 @@
+class Api::V0::ChangesController < Api::V0::ApiController
+  def index
+    query = change_collection
+    paging = pagination(query)
+    changes = query.limit(paging[:page_items]).offset(paging[:offset])
+
+    render json: {
+      links: paging[:links],
+      data: changes.as_json(methods: :current_annotation)
+    }
+  end
+
+  def show
+    render json: {
+      links: {
+        page: api_v0_page_url(page),
+        from_version: api_v0_page_version_url(page, change.from_version),
+        to_version: api_v0_page_version_url(page, change.version)
+      },
+      data: change.as_json(methods: :current_annotation)
+    }
+  end
+
+  protected
+
+  def page
+    return nil unless params.key? :page_id
+    @page ||= Page.find(params[:page_id])
+  end
+
+  def change
+    @change ||=
+      if params[:id].include?('..')
+        from_id, to_id = params[:id].split('..')
+        if from_id.present?
+          Change.between(from: Version.find(from_id), to: Version.find(to_id))
+        else
+          Version.find(to_id).change_from_previous ||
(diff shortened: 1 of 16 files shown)


Mr0grog:
Because we will be adding more numeric things (e.g. `significance`), throwing this in here as a general question on the design of those query params… they currently handle numeric queries like so:

- `[attribute]_gt`/`[attribute]_gte` - filter on `attribute` values that are greater than (/or equal to) the value
- `[attribute]_lt`/`[attribute]_lte` - filter on `attribute` values that are less than (/or equal to) the value

But a couple of alternative approaches:

1. `[attribute]%3E=[value]` / `[attribute]%3E%3D=[value]` - these look a little ugly in the final URL, but using a querystring encoder here lets you basically say: `{ 'priority>=': 0.5 }`. e.g:

    - `priority%3E%3D=0.25&priority%3C%3D=0.5` - 0.25 <= priority <= 0.5
    - `priority%3E%3D=0.25&priority%3C=0.5` - 0.25 <= priority < 0.5
    - `priority%3C%3D=0.5` - priority <= 0.5
    - `priority%3E%3D=0.5` - priority >= 0.5

2. `[attribute]={"gt": [value]}` / `[attribute]={"gt": [value], "lt": [value]}` / `[attribute]={"gte": [value], "lte": [value]}` - send in a JSON object for the value that has fields indicating the bounds of the query. e.g:

    - `priority={"gte": 0.25, "lte": 0.5}` - 0.25 <= priority <= 0.5
    - `priority={"gte": 0.25, "lt": 0.5}` - 0.25 <= priority < 0.5
    - `priority={"gt": 0.25, "lte": 0.5}` - 0.25 < priority <= 0.5
    - `priority={"lte": 0.5}` - priority <= 0.5
    - `priority={"gte": 0.5}` - priority >= 0.5

3. `[attribute]=[gte_value]..[lte_value]` / `[attribute]=[gte_value]...[lt_value]` - similar to how we’ve done date and version ranges. Downside: you can’t do “greater than,” only “greater than or equal to.” e.g:

    - `priority=0.25..0.5` - 0.25 <= priority <= 0.5
    - `priority=0.25...0.5` - 0.25 <= priority < 0.5
    - `priority=..0.5` - priority <= 0.5
    - `priority=0.5..` - priority >= 0.5

   (Though maybe we could get “greater than by post-fixing with a `+`, e.g. `priority=5+..` for “priority > 0.5” ? Getting a little funky and non-standard then, though…)

4. `[attribute]=[math interval notation]` Using the standard mathematical notation, where square brackets indicate “or equal to” and parens otherwise, e.g:

    - `priority=[0.25,0.5]` - 0.25 <= priority <= 0.5
    - `priority=[0.25,0.5)` - 0.25 <= priority < 0.5
    - `priority=(0.25,0.5]` - 0.25 < priority <= 0.5
    - `priority=[,0.5]` - priority <= 0.5
    - `priority=[0.5,]` - priority >= 0.5

lightandluck:
My vote is for 4-math interval notation. It's succinct, readable, no need to send an object (I'm partial to simple GET requests where possible), and seems like it can handle all use cases without being wonky like 3. Really dig it!

app/controllers/api/v0/changes_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    if change
+      api_v0_page_change_url(*args)
+    else
+      api_v0_page_changes_url(*args)
+    end
+  end
+
+  def changes_collection
+    collection = Change
+    where_numeric_range_params(collection, :priority)
```
lightandluck: Will querying on `significance` be added to this PR or a later one?

Mr0grog: Was going to try and merge this first, then add it in #105, since `significance` doesn’t even exist yet in the underlying system here.

lightandluck: Ahh, got it, cool!

Issue #102, opened 2017-07-12
Update changes’ priority attribute based on annotations


Mr0grog:
We originally had the `priority` attribute on `Change` objects as a float between `0` and `1`, indicating the priority of the change for humans to evaluate. We also thought we’d set it by posting an annotation with a specially named field (probably `priority`, but we never really decided). We need to actually put this into practice.

However, now that a lot of time has passed, I think I’ve got some questions or confusion I’d appreciate feedback on:

1. Somewhere along the line this got kinda fuzzy for me and merged with the concept of “significance” in my head. (In fact, I just claimed to @lightandluck that we had a `significance` attribute, but was actually thinking of this `priority` attribute). Are these different concepts or no? Do we need a second `significance` attribute that works similarly or should we merge these two ideas? (Which name do we choose?)

2. I initially wrote this with a default of `0.5`. Annotating with a lower value would effectively deprioritize a change, while a higher value would prioritize. Not sure this was the right decision, especially in the case of “significance.” Should this have a `NULL` or `-1` or similar “unset” value for its default instead? Or just `0` (I feel like we need the ability deprioritize/mark-as-definitely-not-worth-looking-at, so I’m not a huge fan of `0`)?

3. What annotation property name should cause this field to update? (Or should we abandon that idea and just `POST /api/v0/pages/{id}/changes/{a}..{b}/priority`?)

4. Does a second change to the `priority` attribute replace the previous value or does it somehow combine/multiply? (I’m thinking replace, for simplicity’s sake.)

/cc @danielballan @lightandluck 

(Marking as priority because we need this or something related to store significance data in the DB as discussed this afternoon.)

danielballan:
1. I think a Change has both a `priority` and a `significance`. The `priority` will be used for sorting Changes for the hypothetical analyst who only has time to review some of the Pages that they are tasked with. I think our filtering pipeline will seed each Change with an initial priority. (At first, it might be as simple as, "If the bytes are identical, `priority=0`; otherwise `priority=1`. It can grow smoothly for there.) The Change's `significance` is aggregate property of its annotations.

2. I like `NULL` for unset. (`-1` as a sentinel makes it easy to commit math errors.) I agree that we need `0` for "lowest priority" and "lowest significance".

[...]

PR #105, opened 2017-07-14
Change priority/significance cleanup

Mr0grog:
~**Work in progress; DO NOT MERGE**~

This is tracking the work for resolving #102, which is a general cleanup of `priority`, `significance` and how those fields get set. Discussion in the issue is not entirely resolved yet, so this will get worked on as discussion evolves there.


Mr0grog:
Rebased on master, added `significance`.

Only remaining bit is to update `priority` and `significance` fields based on annotations.

Mr0grog:
This now oughta be good to go; will merge shortly if nobody has concerns.

PR #106, opened 2017-07-16
Documented source_metadata for Versionista

michardy:
I don't know if the `/schema` folder is the right place for this especially because it does not link with swagger.


schema/versonista.json
```diff
@@ -0,0 +1,36 @@
+{
```
Mr0grog: Typo in the name of this file!

swagger.yaml
```diff
@@ -183,8 +185,12 @@ paths:
           type: string
         - name: source_metadata
           in: query
-          description: '???'
+          description: >-
+            A JSON blob containing extra source specific information.  This
+            field is free-form, but we generally expect the data will follow
+            one of the formats outlined in the `/schema` folder.
```
Mr0grog: I think it would be helpful if `/schema` were a link to the actual JSON file in question.

michardy: I debated doing this but was not sure where to link it to.  What URL would `/app/assets/schema` map to?

Mr0grog: The local URL would just be `/assets/[file name without the folder name]`, e.g. `<a href="/assets/versionista.json">`

You could also use the absolute version, but then staging would point to production’s docs :\

(that would be `https://api.monitoring.envirodatagov.org/assets/versionista.json`)

michardy: Are you sure that URL works?  My attempts to load existing assets seem to fail:
- https://api.monitoring.envirodatagov.org/assets/swagger.yml (404 & bytes)
- https://api.monitoring.envirodatagov.org/assets/home.js (404 & html)

michardy:
Do we intend on using the [byte array format](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/Data%20from%20different%20sources.md#internet-archive) for internet archive?
If so, it may not be possible to document it in swagger or json schema.

Mr0grog:
> Do we intend on using the byte array format for internet archive? If so, it may not be possible to document it in swagger or json schema.

No—`source_metadata` will _always_ be a JSON object for all source types. I don’t think anybody has sat down to determine what should actually be in the metadata for IA, though: https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/web_monitoring/internetarchive.py#L154

swagger.yaml
Mr0grog: Oh, hrm, in production, you have to have the asset hash on the end (e.g. `https://api.monitoring.envirodatagov.org/assets/swagger-9c4ff43b64cc07ca722a29004093fd3ee86cdf46152aa7eab2f6fd3794f34736.yml`).

Not sure how we’d get that in a static YML file. I guess we could have a URL that serves up those files, but that is starting to feel a little excessive :\

Mr0grog: Oh! Actually, you can rename `swagger.yml` to `swagger.yml.erb` and then have a description field like:

```yml
- description: >-
  Here is a link to
  <a href="<%= asset_path('versionista.json') %>">the Versionista schema</a>.
```

michardy: Will that interfere with Swagger? The server may return HTML rather than a YAML file when swagger queries it.  If Swagger uses GitHub it may instead get an erb file.

Mr0grog: No, it’ll return a YML file if it’s a `.yml.erb`. All that happens is that the file is first run through the ERB processor when compiling assets, then the `.erb` extension is stripped and it is served exactly as it otherwise would have been. Have tested, too :)

Mr0grog:
This is looking really great to me. I am happy to merge in if you there’s nothing you want to add here, @michardy!

michardy:
I have one question.  Should `source_metadata` for versionista be documented in `schema/` and `definitions` as it is currently?

Mr0grog:
If it’s going to be the exact same thing in both, I do feel like the extra source_metadata-specific files in the `/schema` folder are a little redundant.

I think I was assuming initially they were going to be less of a JSONSchema file (which is exactly what the `definitions` section is) and more of a textual document that wouldn’t be compatible with the `definitions` section.

michardy:
If I stick the `source_metadata` definition for versionista in `definitions`, I cannot link to it because swagger's scripts do not generate an HTML id attribute.  This means that we can either, make the docs easy to find by linking them in a separate file, or make the docs on the same page.  If we put them on the same page we should probably say that they can be found at the bottom of the page in the "models" section.

Mr0grog:
Since no other analogous part of the docs is split out into a separate page, I don’t really think we should do any different here unless we have a huge reason to (and it doesn’t sound like we do).

> If we put them on the same page we should probably say that they can be found at the bottom of the page in the "models" section.

Definitely. Where we have links, we should replace them with that text. Similarly, it might be good to have add a `description` to the `source_metadata` field on the `Version` definition that basically says the same thing.

If you want to do the above, I’ll go ahead and merge #110 in now; you’ll want to rebase and clean up the merge conflicts.

michardy:
Ok, merge #110

Mr0grog:
Done.

michardy:
I have removed the link. and updated the pull request to the most recent merge. 

> Similarly, it might be good to have add a `description` to the `source_metadata` field on the `Version` definition that basically says the same thing.

Unfortunately swagger does not let me put a `description` in the `source_metadata` field of `Version`

Mr0grog:
This appears to validate and render for me:

```yml
definitions:
  Version:
    type: object
    properties:
      uuid:
        type: string
        format: uuid4
      source_metadata:
        type: object
        description: >-
          The type of object in this field is based on `source_type`.
          For `source_type = versionista`, this should be a
          `source_metadata_versionista` object.
```

screen shot 2017-07-25 at 7 00 19 pm

…but maybe we’re thinking of different things here?

michardy:
I thought that was that was an undefined behavior because it was not in the docs.  It appears not to be though.

Mr0grog:
Awesome. Good to merge?

michardy:
Yes

PR #108, opened 2017-07-19
HOTFIX: Always query changes based on ID attributes

Mr0grog:
ActiveRecord's `where` method can transparently take an ID or instance for collection attributes, but the `create` and `new` methods cannot. We used to rely on the former behavior, but that would break when new instances were needed. Now we explicitly ensure we have an ID and always use the ID attributes rather than the collection attributes.


Mr0grog:
(Found while working on https://github.com/edgi-govdata-archiving/web-monitoring-ui/pull/82)

PR #114, opened 2017-07-25
Stop exploding when posting empty annotations

### LINKED ISSUE #113: Posting nothing or an empty object for an annotation results in 500 error
The errors are different, but these are similar poor inputs that we are not handling well. These should be 400 errors.

Post nothing at all:

```
JSON::ParserError in Api::V0::AnnotationsController#create
765: unexpected token at ''

app/controllers/api/v0/annotations_controller.rb:33:in `create'
```

Post an empty object (`{}`):

```
app/controllers/api/v0/annotations_controller.rb:33:in `create'
undefined method `change' for nil:NilClass

app/controllers/api/v0/annotations_controller.rb:18:in `show'
app/controllers/api/v0/annotations_controller.rb:37:in `create'
```

As observed in edgi-govdata-archiving/web-monitoring-ui#82

Mr0grog:
Fixes #113. We previously had oddball server errors for sending empty bodies or empty objects as annotations. This makes both of those cases `4xx` errors.

- Empty POST body: 400 (bad request)
- Empty annotation object: 422 (unprocessable entity)

The issue was originally noted in edgi-govdata-archiving/web-monitoring-ui#82. Based on that, I’m wondering if I should actually be accepting empty objects (but not empty bodies). The current code here just formalizes existing behaviors, where we rejected empty objects in one part of the code and then failed to handle it later. Now that POSTing a new annotation *replaces* your previous annotation (see #48), maybe we should allow empty objects.

### DIFF
--- a/app/controllers/api/v0/annotations_controller.rb
+++ b/app/controllers/api/v0/annotations_controller.rb
@@ -33,7 +33,17 @@ class Api::V0::AnnotationsController < Api::V0::ApiController
   end
 
   def create
-    data = JSON.parse(request.body.read)
+    begin
+      data = JSON.parse(request.body.read)
+    rescue JSON::ParserError
+      raise Api::InputError, "Invalid JSON data: '#{request.body.read}'"
+    end
+
+    # TODO: it would be nice to handle this in annotation validation, but we
+    # also want to prevent persisting a change that doesn't exist yet if the
+    # annotation is invalid. Find a better way to do this if possible.
+    raise Api::UnprocessableError, 'Annotations cannot be empty' if data.empty?
+
     @annotation = parent_change.annotate(data, current_user)
     parent_change.save!
     # TODO: should we also somehow return the change's `current_annotation`?
--- a/app/models/annotation.rb
+++ b/app/models/annotation.rb
@@ -40,7 +40,7 @@ class Annotation < ApplicationRecord
     return if value.nil?
 
     if !value.is_a?(Numeric)
-      errors.add(:annotaiton, ".#{property} must be a number")
+      errors.add(:annotation, ".#{property} must be a number")
     elsif !range.cover?(value)
       errors.add(:annotation, ".#{property} must be between #{range.begin} and #{range.end}")
     end
--- a/lib/api/api.rb
+++ b/lib/api/api.rb
@@ -35,6 +35,12 @@ module Api
     end
   end
 
+  class UnprocessableError < ApiError
+    def status_code
+      422
+    end
+  end
+
   class MismatchedHashError < ApiError
     def status_code
       502
--- a/test/controllers/api/v0/annotations_controller_test.rb
+++ b/test/controllers/api/v0/annotations_controller_test.rb
@@ -249,4 +249,34 @@ class Api::V0::AnnotationsControllerTest < ActionDispatch::IntegrationTest
       name: 'Annotations'
     )
   end
+
+  test 'it rejects no POST body' do
+    page = pages(:home_page)
+
+    sign_in users(:alice)
+    post(
+      api_v0_page_change_annotations_path(page, "..#{page.versions[0].uuid}"),
+      headers: { 'Content-Type': 'application/json' },
+      params: ''
+    )
+
+    assert_response :bad_request
+    body = JSON.parse(@response.body)
+    assert(body.key?('errors'), 'Response should have an "errors" property')
+  end
+
+  test 'it rejects an empty annotation (i.e. `{}`)' do
+    page = pages(:home_page)
+
+    sign_in users(:alice)
+    post(
+      api_v0_page_change_annotations_path(page, "..#{page.versions[0].uuid}"),
+      headers: { 'Content-Type': 'application/json' },
+      params: '{}'
+    )
+
+    assert_response :unprocessable_entity
+    body = JSON.parse(@response.body)
+    assert(body.key?('errors'), 'Response should have an "errors" property')
+  end
 end


Mr0grog:
Now that you’re starting to get back up to speed, @danielballan, I’d love any feedback on this behavior. @b5, if you have 15 minutes to think about this, that would be really helpful, too. I don’t think there’s any rush here.

As a review…

When we first added annotations, the contract was that they couldn’t be edited. If a user annotated the same change twice, there would be two annotations (the `Change.current_annotation` property is all the existing annotations merged in chronological order, so not all annotations need have the same fields). We thought this might also be nice for auditing purposes.

When we started thinking about the UI, we thought it might be good not to need an explicit save button and instead save automatically whenever the user interacts with the annotation form. That would result in a *huge* number of annotations on any given change, though. To solve that, we switched to having only one annotation per user, where a creating an annotation replace’s the user’s previous annotation, if any (#48).

In the original design (append-only/non-editable) it didn’t make sense to accept an empty annotation with no data because that would have no effect. However, in fixing this bug, I realized POSTing a non-empty annotation does have a meaningful effect now—because it can replace an existing annotation, it would zero out your previous annotations. It also broke a nice contract we originally had, where an annotation is simply the set of changes you wish to make to the current annotation. Now a user (or the UI serving them) needs to be careful to include all the fields previously used in that user’s annotation alongside all those changes.

A good concrete example of the above is that the analysis process is currently two steps:

1. Users fill out a bunch of checkboxes for various aspects of the change in step one, and
2. In step two they do a more detailed look with tagging and written descriptions of the change.

If the same user did both rounds, the second-round code would have to include all the fields and data about the first round alongside the form data for the second round that the user is actually working with.

I’m feeling like that might be a real problem I overlooked before. Some approaches I’m thinking of:

1. Still make POSTing an annotation edit the user’s existing annotation, but treat it as new data to merge in, rather than replacing the old annotation. This still has a subtle side effect, but maybe not a big deal—if the sequence of annotations is:

    ```js
    // Annotation sequence:
    User A: {field1: 'value1-1', field2: 'value2-1'}
    User B: {field1: 'value1-2',                     field3: 'value3-1'}
    User A: {                                        field3: 'value3-2'}

    expect= {field1: 'value1-2', field2: 'value2-1', field3: 'value3-2'}
    actual= {field1: 'value1-1', field2: 'value2-1', field3: 'value3-2'}
                     ^^^^^^^^^^
    // Actual `current_annotation` has a different `field1` value because
    // editing makes User A’s annotation newer than User B’s
    ```

2. Edit a user’s existing annotation only if it is the newest annotation (and do so by merging instead of replacing, as above), otherwise, if a different user has since added an annotation, create a new annotation. In the above example, you’d wind up with 3 annotations (instead of two) and the problem would be resolved.

3. Just don’t worry about it  ¯\\\_(ツ)\_/¯

4. Something else?

Mr0grog:
Leaning towards (2) above. It's a little more complicated to reintroduce the idea of multiple annotations per user, but setting up that idea lets us split off annotations on other conditions in the future, e.g. tracking the user agent an annotation was made with.

danielballan:
Can the load the current annotation data into its form so that any additional saves will represent changes based on the original state?

Mr0grog:
@danielballan Are you asking if the UI can? Or saying that the DB API here should always treat a saved annotation as the submitted fields + the current merged state of all existing annotations, so in the example above, you’d have…

```js
User A (posted):        {field1: 'value1-1', field2: 'value2-1'}
Result in DB:
    User A:             {field1: 'value1-1', field2: 'value2-1'}
    current_annotation: {field1: 'value1-1', field2: 'value2-1'}

User B (posted):        {field1: 'value1-2',                     field3: 'value3-1'}
Result in DB:
    User A:             {field1: 'value1-1', field2: 'value2-1'}
    User B:             {field1: 'value1-2', field2: 'value2-1', field3: 'value3-1'}
    current_annotation: {field1: 'value1-1', field2: 'value2-1'}

User A (posted):        {                                        field3: 'value3-2'}
Result in DB:
    User B:             {field1: 'value1-2', field2: 'value2-1', field3: 'value3-1'}
    User A:             {field1: 'value1-2', field2: 'value2-1', field3: 'value3-2'}
    current_annotation: {field1: 'value1-2', field2: 'value2-1', field3: 'value3-2'}
```

?

If the former, I think the UI already does this. I wasn’t comfortable placing a requirement on auto-analysis tools or anything else in future needing to load the Change record and its `current_annotation` before posting, though.

If the latter, that would work, but it feels kinda weird to me; we’d no longer really be accurately recording what a someone or some machine actually said about a change (which I kinda felt like was the main reason for keeping `annotation` records around instead of just modifying `current_annotation`).

danielballan:
I see where the confusion is. I imagined that each unique would start with a blank slate, with no annotation. Users could update their own annotations, but annotations from different users would be non-interfering.

Some admin view would present _all_ the annotations.

The idea of a "current_annotation" has always seemed muddy to me, but I don't think it's possible in general to merge annotations together.

Mr0grog:
Ha. I feel like `current_annotation` is the one that is primarily useful, while the full list of annotation is more of an audit tool.

e.g. Tom does a first-round analysis and checks off some boxes and makes some simple notes, then marks as 0.5 significance; Jane and Alice then each do a second-round deeper analysis (if it’s worth the deeper look, it might be worth two pairs of eyes)—they each add some tags, maybe one corrects an ealier checkbox that was mischecked, etc.

If you want to know what we as a team have to say about that page, you’d want to check the `current_annotation`. Each individual annotation only has a piece of the picture.

danielballan:
Got it. I think my concept came from imagining (prematurely) the day when we could have a horde of semi-trusted volunteer contributors working independently, not a small stable of trusted volunteers coordinating closely.

danielballan:
I don't see an obviously-correct approach here. I will think on this tonight.

gblike:
I don't entirely follow all the annotations business logic debate, but I do like better error handling for empty posts. I feel like having an empty post should not exist. :ship:

Mr0grog:
Yeah, I’m sure this is at least partially confusing because the code doesn’t match up with the discussion here at all. Proposing this PR kicked off a whole different discussion that then wound up here and we wanted to wait to resolve it before merging. (Actually, we did resolve it and @danielballan was going to implement on top of this PR, but he never got around to it.)

That said, the code here is probably reasonable to merge without resolving that discussion. It doesn’t really change any behavior except the error codes you get—4XX instead of 500, so maybe I should just rebase and merge it.

All the discussion about annotation sequencing and mutability here is the middle of a much longer-running conversation with more that came both before and after this (and simultaneously out of band on phone calls). It probably requires a little more context (which I’m happy to walk you through, if you want).

weatherpattern:
pardon the interruption... @Mr0grog has mentioned walk through of the db repo a couple of times, here and on the last call. Would that be beneficial? I'd be interested in participating, and I think it would be useful to record it, so that we have record of some of the thinking.

Mr0grog:
OK, I think I need to move all the conversation about logic around saving, updating, and ordering annotations to a new issue, then fix the conflicts in this code so it can be merged. We got really off-topic here (on important things!), but since annotations are coming back to life again, we should get back to making sure stuff like this is cleaned up.

Mr0grog:
Conversation moved to #375.

Code here is rebased and should be good to go; people already sorta said this was good, but will wait until tonight if anyone wants to take a look since this is OLD.

PR #126, opened 2017-08-20
HOTFIX: Reflect origin header for CORS auth

Mr0grog:
Looks like I broke remote authentication in e70230a. An upgrade to rack-cors changed the behavior of `origins '*'` such that it now does not reflect the request's origin and instead actually sends '*' for the allowed origins header. That means authentication on our requests fails :(

See this change in rack-cors: cyu/rack-cors#142

We needed this behavior initially because it seemed like `Authorization` headers in our requests were getting stripped, but it seems like this may no longer be true (or, more likely, I mis-diagnosed the issue originally). That needs more testing, though. For now, force rack-cors to return to its earlier behavior.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -24,7 +24,7 @@ module WebpageVersionsDb
       end
 
       allow do
-        origins '*'
+        origins /.*/
         resource '*', :headers => :any, :methods => [:get, :post, :options], :if => is_admin_url
       end
     end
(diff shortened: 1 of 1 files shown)


Mr0grog:
Side note for the curious: we largely work around the associated vulnerability here by disallowing cookie data on all remote requests.

PR #127, opened 2017-08-21
Convert `errorCode` -> `error_code` in `source_metadata`

Mr0grog:
Really not sure how this error ever made it in and how I let it sit for so long without addressing it :(

See also: edgi-govdata-archiving/web-monitoring-versionista-scraper#39


Mr0grog:
Rebased on master. Versionista scraper and the new IA import code in `-processing` both are creating new records with `error_code` now, so I think it’s time to merge and do this cleanup migration.

PR #130, opened 2017-09-09
Sync Page title with most recently captured version

### LINKED ISSUE #59: Track titles on Version instead of Page
See #57 for the original discussion on this.

Since page titles can change over time (and are really just part of a page’s content), we should add a `title` field to the Version model and make Page’s `title` merely a reflection of the latest Version’s title.

This can probably be best handled in the [`after_save` callback](http://guides.rubyonrails.org/active_record_callbacks.html#updating-an-object) on Version or the [`after_add` callback](http://guides.rubyonrails.org/association_basics.html#association-callbacks) on Page’s association with versions. We should also keep in mind that the Page’s title should always reflect the version with the latest `capture_time`, not necessarily the last one added to the database (they can be added out of order).

maxtedford:
Resolves #59 

This work adds a `title` attribute to the `versions` table, and supplies an `after_add` callback to `Page` model, which sets the `Page`'s `title` attribute to match that of the `Version` with the most recent `capture_time`.

I'm not quite familiar enough with the application at this point to know intuitively whether the `title` attribute on a `Version` should include a validation for presence as a mandatory field, but my best judgement led me to forego that addition, and simpy match it to the `Page`'s implementation of the same field.

Please let me know if I've missed anything or if there's anything else I can do to help on this ticket! 😄

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -1,7 +1,11 @@
 class Page < ApplicationRecord
   include UuidPrimaryKey
 
-  has_many :versions, -> { order(capture_time: :desc) }, foreign_key: 'page_uuid', inverse_of: :page
+  has_many :versions,
+    -> { order(capture_time: :desc) },
+    foreign_key: 'page_uuid',
+    inverse_of: :page,
+    after_add: :sync_titles
   has_one :latest,
     (lambda do
       # DISTINCT ON requires the first ORDER to be the distinct column(s)
@@ -36,6 +40,13 @@ class Page < ApplicationRecord
   def url_must_have_domain
     unless url.match?(/^([\w\+\-\.]+:\/\/)?[^\/]+\.[^\/]{2,}/)
       errors.add(:url, 'must have a domain')
+    end
+  end
+
+  def sync_titles(version)
+    if version.title.present?
+      most_recent_capture_time = versions.limit(1).order(capture_time: :desc).pluck(:capture_time).first
+      self.title = version.title if most_recent_capture_time.nil? || most_recent_capture_time <= version.capture_time
     end
   end
 end
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -45,8 +45,8 @@ class Page < ApplicationRecord
 
   def sync_titles(version)
     if version.title.present?
-      most_recent_capture_time = versions.limit(1).order(capture_time: :desc).pluck(:capture_time).first
-      self.title = version.title if most_recent_capture_time.nil? || most_recent_capture_time <= version.capture_time
+      most_recent_capture_time = latest.capture_time
+      update(title: version.title) if most_recent_capture_time.nil? || most_recent_capture_time <= version.capture_time
     end
   end
 end
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -4,8 +4,7 @@ class Page < ApplicationRecord
   has_many :versions,
     -> { order(capture_time: :desc) },
     foreign_key: 'page_uuid',
-    inverse_of: :page,
-    after_add: :sync_titles
+    inverse_of: :page
   has_one :latest,
     (lambda do
       # DISTINCT ON requires the first ORDER to be the distinct column(s)
@@ -40,13 +39,6 @@ class Page < ApplicationRecord
   def url_must_have_domain
     unless url.match?(/^([\w\+\-\.]+:\/\/)?[^\/]+\.[^\/]{2,}/)
       errors.add(:url, 'must have a domain')
-    end
-  end
-
-  def sync_titles(version)
-    if version.title.present?
-      most_recent_capture_time = latest.capture_time
-      update(title: version.title) if most_recent_capture_time.nil? || most_recent_capture_time <= version.capture_time
     end
   end
 end
--- a/app/models/version.rb
+++ b/app/models/version.rb
@@ -8,7 +8,7 @@ class Version < ApplicationRecord
     class_name: 'Change',
     foreign_key: 'uuid_to'
 
-  after_save :sync_page_title
+  after_create :sync_page_title
 
   def previous
     self.page.versions.where('capture_time < ?', self.capture_time).first
--- a/app/models/version.rb
+++ b/app/models/version.rb
@@ -8,11 +8,22 @@ class Version < ApplicationRecord
     class_name: 'Change',
     foreign_key: 'uuid_to'
 
+  after_save :sync_page_title
+
   def previous
     self.page.versions.where('capture_time < ?', self.capture_time).first
   end
 
   def change_from_previous
     Change.between(from: previous, to: self)
+  end
+
+  private
+
+  def sync_page_title
+    if title.present?
+      most_recent_capture_time = page.latest.capture_time
+      page.update(title: title) if most_recent_capture_time.nil? || most_recent_capture_time <= capture_time
+    end
   end
 end
--- a/db/migrate/20170909151007_add_title_to_versions.rb
+++ b/db/migrate/20170909151007_add_title_to_versions.rb
@@ -1,0 +1,5 @@
+class AddTitleToVersions < ActiveRecord::Migration[5.1]
+  def change
+    add_column :versions, :title, :string
+  end
+end
--- a/db/migrate/20170909151007_add_title_to_versions.rb
+++ b/db/migrate/20170909151007_add_title_to_versions.rb
@@ -1,5 +1,8 @@
 class AddTitleToVersions < ActiveRecord::Migration[5.1]
   def change
     add_column :versions, :title, :string
+    Version.includes(:page).find_each do |v|
+      puts "Updated Version record with UUID #{v.uuid} to have title '#{v.title}'" if v.update(title: v.page.title)
+    end
   end
 end
--- a/db/migrate/20170909151007_add_title_to_versions.rb
+++ b/db/migrate/20170909151007_add_title_to_versions.rb
@@ -1,8 +1,20 @@
 class AddTitleToVersions < ActiveRecord::Migration[5.1]
   def change
     add_column :versions, :title, :string
-    Version.includes(:page).find_each do |v|
-      puts "Updated Version record with UUID #{v.uuid} to have title '#{v.title}'" if v.update(title: v.page.title)
+
+    reversible do |dir|
+      dir.up do
+        say_with_time("Copying titles from pages to versions") do
+          page_count = 0
+          version_count = 0
+          Page.includes(:versions).find_each do |page|
+            page_count += 1
+            version_count += page.versions.update_all(title: page.title)
+          end
+          say("#{page_count} pages assessed", :subitem)
+          say("#{version_count} versions updated", :subitem)
+        end
+      end
     end
   end
 end
--- a/db/migrate/20170909151007_add_title_to_versions.rb
+++ b/db/migrate/20170909151007_add_title_to_versions.rb
@@ -4,7 +4,7 @@ class AddTitleToVersions < ActiveRecord::Migration[5.1]
 
     reversible do |dir|
       dir.up do
-        say_with_time("Copying titles from pages to versions") do
+        say_with_time('Copying titles from pages to versions') do
           page_count = 0
           version_count = 0
           Page.includes(:versions).find_each do |page|
--- a/test/controllers/api/v0/imports_controller_test.rb
+++ b/test/controllers/api/v0/imports_controller_test.rb
@@ -20,7 +20,7 @@ class Api::V0::ImportsControllerTest < ActionDispatch::IntegrationTest
     import_data = [
       {
         page_url: 'http://testsite.com/',
-        page_title: 'Example Page',
+        title: 'Example Page',
         site_agency: 'The Federal Example Agency',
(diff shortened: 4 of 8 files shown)


Mr0grog:
This sounds great, thanks so much @maxtedford! I am super busy this afternoon, but will try and look at this with a closer eye tonight or tomorrow at the latest.

app/models/page.rb
```diff
@@ -38,4 +42,11 @@ def url_must_have_domain
       errors.add(:url, 'must have a domain')
     end
   end
+
+  def sync_titles(version)
+    if version.title.present?
+      most_recent_capture_time = versions.limit(1).order(capture_time: :desc).pluck(:capture_time).first
+      self.title = version.title if most_recent_capture_time.nil? || most_recent_capture_time <= version.capture_time
```
Mr0grog: You need to also call `save!` here to save this change to the DB. You could also call `update` instead of `title=` (`update` immediately saves):

```rb
update(title: version.title) if most_recent_capture_time.nil? || most_recent_capture_time <= version.capture_time
```

See below comment in the tests about making sure the changes are saved, too.

app/models/page.rb
```diff
@@ -38,4 +42,11 @@ def url_must_have_domain
       errors.add(:url, 'must have a domain')
     end
   end
+
+  def sync_titles(version)
+    if version.title.present?
+      most_recent_capture_time = versions.limit(1).order(capture_time: :desc).pluck(:capture_time).first
```
Mr0grog: It might be easier/clearer to just call `latest` here:

```rb
most_recent_capture_time = latest.capture_time
```

db/migrate/20170909151007_add_title_to_versions.rb
```diff
@@ -0,0 +1,5 @@
+class AddTitleToVersions < ActiveRecord::Migration[5.1]
+  def change
+    add_column :versions, :title, :string
```
Mr0grog: I think we should also make sure we fill in the `title` attribute on all existing versions with the existing title from their page. You could do that here or make a second migration for it.

test/models/page_test.rb
```diff
@@ -13,4 +13,24 @@ class PageTest < ActiveSupport::TestCase
     page = Page.create(url: 'some/path/to/a/page')
     assert_not(page.valid?, 'The page should be invalid because it has no domain')
   end
+
+  test 'page title should sync with title from version with most recent capture time' do
+    page = pages(:home_page)
+    assert_equal('Page One', page.title)
+
+    page.versions.create(title: 'Newest Version', capture_time: '2017-03-05T00:00:00Z')
+    assert_equal('Newest Version', page.title, 'The page title should always sync against the title of the version with the most recent capture time')
```
Mr0grog: Alongside this assertion, you also need to make sure the changes were persisted to the DB (see note in the model file). You should get a fresh copy of the page from the DB to make assertions on:

```rb
page.versions.create(title: 'Newest Version', capture_time: '2017-03-05T00:00:00Z')

# Get a fresh page instance
page = Page.find(page.uuid) # OR: page.reload
assert_equal('Newest Version', page.title, '...')
# etc.
```

You could also check that the page wasn’t left with unsaved changes:

```rb
assert_not(page.changed?, 'The page was left with unsaved changes')
```

test/models/page_test.rb
```diff
@@ -13,4 +13,24 @@ class PageTest < ActiveSupport::TestCase
...
+    page.versions.create(title: 'Older Version', capture_time: '2017-03-01T00:00:00Z')
+    refute_equal('Older Version', page.title, 'The page title should not sync against the title of a newly created version with an older capture time')
+  end
+
+  test "page title should not sync with a version's title if it's nil or blank" do
+    page = pages(:home_page)
+    assert_equal('Page One', page.title)
+
+    page.versions.create(capture_time: '2017-03-05T00:00:00Z')
+    assert_equal('Page One', page.title, 'The page title should not sync with the incoming version if it has a nil title')
```
Mr0grog: Same issue about page saving as above.

Mr0grog:
Also, pro tip: if you put “fixes #59” (or “closes” or “resolves”) instead of “delivers,” GitHub will automatically close the issue when this is merged! It’s super convenient. More info: https://help.github.com/articles/closing-issues-using-keywords/

maxtedford:
Hey @Mr0grog, thanks for the thorough review! I'm totally embarrassed that I didn't actually save the record as a part of the `after_add` callback. I think in my head I was treating it as a `before_action`, and thus assumed an automatic save as a part of the process. But you're absolutely right, so I updated it accordingly, including the tests.

I didn't even notice the `latest` relation you had on the `Page` model. That's super helpful! As I was making updates to the `import_versions_job` to no longer create new pages with a `title` attribute, I considered adding a method alias to simply pull the `title` from the `latest` record (something like the below)

```ruby
alias_method :original_title, :title

def title
end
```

, but decided instead to simply update the test to `assert_nil pages[0].title` and `assert_equal [

db/migrate/20170909151007_add_title_to_versions.rb
maxtedford: Makes sense to me. I just went ahead and added that to this existing migration.

db/migrate/20170909151007_add_title_to_versions.rb
```diff
@@ -0,0 +1,8 @@
...
+    Version.includes(:page).find_each do |v|
+      puts "Updated Version record with UUID #{v.uuid} to have title '#{v.title}'" if v.update(title: v.page.title)
```
Mr0grog: I think this is going to be verbose enough to be a problem (we have ~335,000 versions in production right now). This much logging will actually slow the migration way down and probably not be helpful. Maybe better to count the number of changes and just output a summary, e.g. “Updated title on X version records for Y pages.”

Also, this should probably only run when migrating up, not down, so maybe move this part into a `reversible` block with only an `up` section?

Finally, Rails has some nice logging helpers for migrations. You could do something like:

```rb
say_with_time("Copying titles from pages to versions") do
  Version.includes(:page).find_each do |v|
    # Update version and count it
  end
  say("X versions", :subitem)
  say("Y pages", :subitem)
end
```

…and you get nice output like:

```
== 20170909151007 AddTitleToVersions: migrating ===============================
-- add_column(:versions, :title, :string)
   -> 0.0013s
-- Copying titles from pages to versions
   -> 1019 versions
   -> 146 pages
   -> 2.3985s
== 20170909151007 AddTitleToVersions: migrated (2.3999s) ======================
```

test/controllers/api/v0/imports_controller_test.rb
```diff
@@ -64,9 +64,10 @@ def teardown
 
     pages = Page.where(url: 'http://testsite.com/')
     assert_equal 1, pages.length
-    assert_equal import_data[0][:page_title], pages[0].title
     assert_equal import_data[0][:site_agency], pages[0].agency
     assert_equal import_data[0][:site_name], pages[0].site
+    assert_equal import_data[0][:page_title], pages[0].latest.title
+    assert_nil pages[0].title
```
Mr0grog: Shouldn’t this *not* be `nil`? i.e. the earlier assertion (`assert_equal import_data[0][:page_title], pages[0].title`) should still be here, shouldn’t it?

db/migrate/20170909151007_add_title_to_versions.rb
Mr0grog: Oh, also: this might be faster if you roll through the list of pages and call `update_all` on their versions; that way there’s only one SQL query per page, rather than one for every version, e.g:

```rb
Page.each do |page|
  page.versions.update_all(title: page.title)
end
```

`update_all` also returns the number of records that were affected, which is helpful in doing the summary counting I mentioned above.

test/models/page_test.rb
```diff
@@ -13,4 +13,32 @@ class PageTest < ActiveSupport::TestCase
...
+  end
+
+  test "page title should not sync with a version's title if it's nil or blank" do
+    page = pages(:home_page)
+    assert_equal('Page One', page.title)
+
+    page.versions.create(capture_time: '2017-03-05T00:00:00Z')
+    page.reload
+    assert_equal('Page One', page.title, 'The page title should not sync with the incoming version if it has a nil title')
+    refute(page.changed?, 'The page was left with unsaved changes')
```
Mr0grog: …I *think* this is pointless after you’ve called `reload` (it trashes all state on the model object). Not totally sure on that, though.

test/controllers/api/v0/imports_controller_test.rb
Mr0grog: We should probably also change the import data here to `title` instead of `page_title`, or have tests that use both (since we’re transitioning from one to the other).

Mr0grog:
> I considered adding a method alias on the Page model to simply pull the title from the latest record in the event of a nil title (something like the below)… but decided instead to simply stick with what you recommended

I think that was a good choice. The virtual attribute would have been convenient, but potentially a big problem for large queries that return many pages (you don’t want those to wind up issuing an additional SQL query for each record). It probably also would have required changes everywhere we JSON serialize pages.

> `assert_nil pages[0].title`

Noted this inline, too, but I’m not sure this makes sense. Shouldn’t the pages still be acquiring the title from its versions? So this shouldn’t be `nil` when all is said and done.

db/migrate/20170909151007_add_title_to_versions.rb
maxtedford: Very cool! I've never made use of these Rails helpers before, so TIL. I'll change that up.

test/controllers/api/v0/imports_controller_test.rb
maxtedford: Yes, it makes sense that this test should still pass as is. I dug a little more into this, and it seems that [this line](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/jobs/import_versions_job.rb#L69) wasn't actually triggering the `after_add` callback on the `Page` model. Interestingly enough, if I move the callback into the `Version` model in the form of an `after_save`, it does get triggered, so I'll push up that change as well.

maxtedford: I'll also revert the test back. I should have trusted the test as it was originally written.

test/models/page_test.rb
maxtedford: I also looked into this, and you're right. The `reload` method _does_ in fact trash any existing state on the model object. Also interestingly, however, I discovered (after throwing a couple `binding.pry` calls between a few of these assertions) that we don't actually need to call `reload` to get the new state on these objects.

```ruby
binding.pry
page.versions.create(title: 'Newest Version', capture_time: '2017-03-05T00:00:00Z')
binding.pry
```

`page.changed?` returned `false` from inside both of those `binding.pry`s despite the `page` record having actually had it's `title` attribute updated. So, it would seem we don't need to worry about calling `reload` to confirm these changes.

maxtedford:
Hey @Mr0grog, I just pushed up a few more small changes based on your latest feedback. Please let me know if there's anything else I can do to help out!

test/controllers/api/v0/imports_controller_test.rb
Mr0grog: Oh, I think the more immediate problem was actually https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/jobs/import_versions_job.rb#L52, which should have been `page.versions.push(version)`

test/models/page_test.rb
Mr0grog: I think this is because you changed the callback to use `update`, which saves immediately, so you *should* get `page.changed? == false`. I definitely tested this with an assertion before you made the latest changes and it did in fact fail. I think it would be good to have that assertion to guard against any regression there.

app/models/version.rb
```diff
@@ -8,11 +8,22 @@ class Version < ApplicationRecord
     class_name: 'Change',
     foreign_key: 'uuid_to'
 
+  after_save :sync_page_title
```
Mr0grog: I’m a little worried that this will be a lot higher overhead than `after_add` on page association, but it’s probably fine. We could do `after_create`, though.

test/controllers/api/v0/imports_controller_test.rb
maxtedford: Ah, yep I think you're right!

app/models/version.rb
maxtedford: Yeah, it probably will seeing as versions are saved in more contexts than just when they're added to a page. I'm obviously happy to go with whichever you think is best, either `after_save` or `after_create`. For the time being, I'll change it to `after_create`, but let me know if you prefer I leave it be.

test/models/page_test.rb
maxtedford: Makes sense to me. I can add it back in as a regression safeguard 👍

maxtedford:
Okay, I've restored the `refute(page.changed?)` to those page model tests as a safeguard, changed the `after_save` on the `Version` model to an `after_create`, and made that offering to the 🤖 👮 gods. Let me know if you notice anything else! 🤞

Mr0grog:
Beautiful, thank you so much for working all the way through this! 🎉 💃

PR #132, opened 2017-09-09
Do not include latest version on pages by default

### LINKED ISSUE #131: Do not include latest version on each page by default
It seems like this query is probably a little expensive, so we might be able to improve performance for simple operations, like just listing versions, by making the inclusion of the latest version an opt-in, similar to including the list of versions. See #128 for more discussion.

Let’s start with a `?include_latest` flag to get the current behavior. I think it should conflict with `?include_versions`, so if you have both, `include_versions` wins. (Another option might be making the versions flag require a value, i.e: `?include_versions=true` and `?include_versions=latest`.)

Mr0grog:
You must now use an `?include_latest` flag:

`GET /api/v0/pages?include_latest`

Note that if you use both `?include_latest&include_versions`, then `include_latest` is ignored.

Fixes #131.

### DIFF
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -27,8 +27,10 @@ class Api::V0::PagesController < Api::V0::ApiController
           .includes(:versions)
           .order('versions.capture_time')
         results.as_json(include: :versions)
-      else
+      elsif should_include_latest
         pages.includes(:latest).as_json(include: :latest)
+      else
+        pages.as_json
       end
 
     render json: {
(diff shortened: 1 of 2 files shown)


app/controllers/api/v0/pages_controller.rb
```diff
@@ -27,8 +27,10 @@ def index
           .includes(:versions)
           .order('versions.capture_time')
         results.as_json(include: :versions)
-      else
+      elsif should_include_latest
```
danielballan: I will never get used to this feature of Ruby syntax. Is it really that hard to type 'e'? End of rant.

Mr0grog: This, coming from a Python person, where that 's' is too hard to type, too? ;)

PR #133, opened 2017-09-13
Rename pagination parameter to 'chunk' and add 'chunk_size' knob.

### LINKED ISSUE #44: Make paging page size customizable
Right now you can ask for a given page of results (with the `?page={number}` querystring arg), but you can’t alter how many items are on that page. We might also want to move away from “page” terminology, too, since that can be confusing with the actual web pages we are tracking.

danielballan:
Closes #44

### DIFF
--- a/app/controllers/concerns/paging_concern.rb
+++ b/app/controllers/concerns/paging_concern.rb
@@ -1,7 +1,8 @@
 module PagingConcern
   extend ActiveSupport::Concern
 
-  PAGE_SIZE = 100
+  DEFAULT_PAGE_SIZE = 100
+  MAX_PAGE_SIZE = 10_000
 
   protected
 
@@ -28,20 +29,21 @@ module PagingConcern
 
     format_type = url_format || self.paging_url_format
     total_items = collection.count
-    total_pages = total_items.zero? ? 1 : (total_items / PAGE_SIZE.to_f).ceil
-    page_number = (params[:page] || 1).to_i.clamp(1, total_pages)
-    page_offset = (page_number - 1) * PAGE_SIZE
+    page_size = (params[:chunk_size] || DEFAULT_PAGE_SIZE).to_i.clamp(1, MAX_PAGE_SIZE)
+    total_pages = total_items.zero? ? 1 : (total_items / page_size.to_f).ceil
+    page_number = (params[:chunk] || 1).to_i.clamp(1, total_pages)
+    page_offset = (page_number - 1) * page_size
 
     links = {
       first: path_resolver.call(
         collection_type,
         format: format_type,
-        params: request.query_parameters.merge(page: 1)
+        params: request.query_parameters.merge(chunk: 1)
       ),
       last: path_resolver.call(
         collection_type,
         format: format_type,
-        params: request.query_parameters.merge(page: total_pages)
+        params: request.query_parameters.merge(chunk: total_pages)
       ),
       prev: nil,
       next: nil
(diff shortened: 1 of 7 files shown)


danielballan:
Tests pass locally. I hope Circle-CI concurs. When I was a lad we used rspec; rubucop is new to me.

danielballan:
I know understand/recall that rubocop is a linter. Catching up...

danielballan:
Currently failing on:

```
Failure:
Api::V0::PagesControllerTest#test_should_repect_chunk_size_pagination_parameter [/Users/dallan/Documents/Repos/web-monitoring-db/test/controllers/api/v0/pages_controller_test.rb:27]:
Expected: 1
  Actual: 4
```

Will revisit tomorrow morning.

danielballan:
🛌  + ☕️  = ✅

app/controllers/concerns/paging_concern.rb
```diff
@@ -28,20 +29,21 @@ def pagination(collection, path_resolver: :paging_path_for, url_format: nil)
 
     format_type = url_format || self.paging_url_format
     total_items = collection.count
-    total_pages = total_items.zero? ? 1 : (total_items / PAGE_SIZE.to_f).ceil
-    page_number = (params[:page] || 1).to_i.clamp(1, total_pages)
-    page_offset = (page_number - 1) * PAGE_SIZE
+    page_size = (params[:chunk_size] || DEFAULT_PAGE_SIZE).to_i.clamp(1, MAX_PAGE_SIZE)
+    total_pages = total_items.zero? ? 1 : (total_items / page_size.to_f).ceil
+    page_number = (params[:chunk] || 1).to_i.clamp(1, total_pages)
+    page_offset = (page_number - 1) * page_size
```
Mr0grog: I know it’ll be more churn, but should we rename these local variables (and the module constants at the top of the file) to be about `chunk` instead of `page`?

test/controllers/api/v0/pages_controller_test.rb
```diff
@@ -11,12 +11,30 @@ class Api::V0::PagesControllerTest < ActionDispatch::IntegrationTest
   end
 
   # Regression
-  test 'should not include the `page` parameter multiple times in one paging link' do
-    # This error occurred when the requested URL already had a `page` param
-    get('/api/v0/pages?page=1')
+  test 'should not include the `chunk` parameter multiple times in one paging link' do
+    # This error occurred when the requested URL already had a `chunk` param
+    get('/api/v0/pages?chunk=1')
```
Mr0grog: This is fine, but two additional useful things you can do here:

- You can generate the URL or path with `api_v0_pages_path` or `api_v0_pages_url`.
- You can incorporate the query params in two ways:
    - An option on the `get` method:

        ```rb
        get(api_v0_pages_path, params: {chunk: 1})
        # or if you know it won’t collide with options that affect how the GET is done:
        get(api_v0_pages_path, chunk: 1)
        ```

    - As an option on the `api_v0_pages_path` or `_url` helper:

        ```rb
        get(api_v0_pages_path(params: {chunk: 1})
        # or if you know it won’t collide with more meaningful options:
        get(api_v0_pages_path(chunk: 1))
        ```

Neither of those are really a big deal, but I think they make the tests a little nicer because you can divorce the method being tested from its actual URL and you can avoid any URL encoding (not that any needs to be done for this particular argument and value).

app/controllers/concerns/paging_concern.rb
danielballan: Sure. I was being uncharacteristically gun shy. I'm a fan of internal consistency in local var names.

test/controllers/api/v0/pages_controller_test.rb
danielballan: :+1:

danielballan:
New commits address review comments.

Mr0grog:
Docs!

danielballan:
Docs updated.

PR #134, opened 2017-09-13
Add EDGI S3 domains to ALLOWED_ARCHIVE_HOSTS.

danielballan:
At some point we discussed adding this to ``.env.example``. I don't think
we came to a clear conclusion, but it seems reasonable to me to add it.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,7 +24,7 @@ export MAIL_SENDER='some-email-account@example.com'
 # export MAIL_SMTP_TLS='true'
 
 # Controls URLs that won't be downloaded and re-hosted when importing versions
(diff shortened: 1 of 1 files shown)


danielballan:
Nah, let's do it right.

Mr0grog:
Well! I don’t know if auto-equating those would be “right.” In rebuttal to myself, not sure more code and special cases is better than a little more wordy basic configuration.

danielballan:
Then I'm inclined to merge and revisit this after v0. Good enough?

Mr0grog:
Agree! Sorry for muddying the waters a bit.

danielballan:
:+1:

PR #135, opened 2017-09-13
Sort versions in /pages?include_versions in descending capture_time order.

### LINKED ISSUE #118: /v0/pages?include_versions returns versions in wrong order
The `/api/v0/pages?include_versions` endpoint lists versions in ascending date order, but all other places where we get a list of versions are in descending order:

- `/api/v0/pages/{id}` The `.versions` property is in descending date order
- `/api/v0/pages/{id}/versions` Records are returned in descending date order

Whatever order we choose, we should be consistent. The descending order versions were intentionally descending and I’m pretty sure I wasn’t thinking about it at all when I added `?include_versions`, so I think that one should be changed to descending order as well.

Examples:

- https://api.monitoring.envirodatagov.org/api/v0/pages?url=https://www.blm.gov/about/how-we-manage/climate-change&include_versions
- https://api.monitoring.envirodatagov.org/api/v0/pages/7a051ddb-dcab-4e24-9b00-cc36cb6a1f3d
- https://api.monitoring.envirodatagov.org/api/v0/pages/7a051ddb-dcab-4e24-9b00-cc36cb6a1f3d/versions

danielballan:
Closes #118

### DIFF
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -25,7 +25,7 @@ class Api::V0::PagesController < Api::V0::ApiController
         results = query
           .where(uuid: page_ids)
           .includes(:versions)
-          .order('versions.capture_time')
+          .order('versions.capture_time DESC')
         results.as_json(include: :versions)
       elsif should_include_latest
         pages.includes(:latest).as_json(include: :latest)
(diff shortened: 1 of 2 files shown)


Mr0grog:
Looks like beautiful wonderfulness

PR #138, opened 2017-09-14
Fill `diff_service_version` from `diff.version`

### LINKED ISSUE #137: Fill in `diff_service_version` with data from differ
We should fill in the [`diff_service_version` field](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/controllers/api/v0/diff_controller.rb#L11) with data from the actual diff response, now that it’s being added there (edgi-govdata-archiving/web-monitoring-processing#89).

Related questions:

- Since we will be copying the `version` field from the response data up one level is the structure of the result, should we remove it from the `content` field? i.e. instead of:

    ```js
    {
      "diff_service": "xxx",
      "diff_service_version": "yyy",
      "content": {
        "version": "yyy" // duplicated info
        "diff": "zzz"
      }
    }
    ```

    We do:

    ```js
    {
      "diff_service": "xxx",
      "diff_service_version": "yyy",
      "content": {
        "diff": "zzz"
      }
    }
    ```

- If yes to the above, should we change to unbox the diff entirely? i.e. instead of:

    ```js
    {
      "diff_service": "xxx",
      "diff_service_version": "yyy",
      "content": {
        "diff": "zzz"
      }
    }
    ```

    We do:

    ```js
    {
      "diff_service": "xxx",
      "diff_service_version": "yyy",
      "content": "zzz"
    }
    ```

- If yes to the above, should we also have a way to do the same if there is other new metadata added? e.g:

    ```js
    {
      "diff_service": "xxx",
      "diff_service_version": "?",
      "content": {
        "diff": "zzz",
        "version": "yyy",
        "something_else": "abc",
        "extra_metadata": "def"
      }
    }
    ```

    to:

    ```js
    {
      "diff_service": "xxx",
      "diff_service_version": "yyy",
      "content": "zzz",
      "metadata": {
        "something_else": "abc",
        "extra_metadata": "def"
      }
    }
    ```

/cc @danielballan

Mr0grog:
Fixes #137.

Note this doesn't address any of the lingering questions from #137. We should work those out and, if needed, add some more commits here before merging.

### DIFF
--- a/app/controllers/api/v0/diff_controller.rb
+++ b/app/controllers/api/v0/diff_controller.rb
@@ -2,14 +2,16 @@ class Api::V0::DiffController < Api::V0::ApiController
   def show
     ensure_diffable
 
+    content = raw_diff
+
     render json: {
       data: {
         page_id: change.version.page.uuid,
         from_version_id: change.from_version.uuid,
         to_version_id: change.version.uuid,
         diff_service: params[:type],
-        diff_service_version: '?',
-        content: raw_diff
+        diff_service_version: content.is_a?(Hash) && content['version'] || '?',
+        content: content
       }
     }
   end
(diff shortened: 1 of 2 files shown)


Mr0grog:
Not going to do this; see discussion in #137.

PR #141, opened 2017-09-22
Update gems; fixe Nokogiri vulnerability

Mr0grog:
More info on vulnerability (fixed in v1.8.1): https://github.com/sparklemotion/nokogiri/issues/1673

This also updates lots of other assorted gems that were a little behind, but doesn't touch some that have had major revisions (e.g. JWT, which I still need to look into and upgrade).


Mr0grog:
Also don’t update Rubocop, which now has new violations. See #142.

PR #147, opened 2017-10-02
Fixes #99 - Integrating CircleCI 2.0

### LINKED ISSUE #99: Upgrade to CircleCI 2.0
CircleCI v2 is out of beta, so we should upgrade when we have time.

https://circleci.com/docs/2.0/migrating-from-1-2/

frederikspang:
Adds integration and removes old Circle 1.0 files, in favour of CircleCI 2.0, as pointed out in #99


shapironick:
we got 99 problems and @frederikspang fixed one :) Thanks Frederik!

.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
...
+      - image: circleci/ruby:2.4.1-node
+        environment:
+          PGHOST: 127.0.0.1
+          PGUSER: root
+          HOST_URL: 'web-monitoring-db.test'
+          ALLOWED_ARCHIVE_HOSTS: 'https://edgi-versionista-archive.s3.amazonaws.com/ https://test-bucket.s3.amazonaws.com/'
+      - image: circleci/postgres:9.5-alpine
+        environment:
+          POSTGRES_USER: root
+          POSTGRES_DB: circle-test_test
```
Mr0grog: Do we actually need either of these environment vars for setting up PG? Since we are manually running `rake db:create` below, we shouldn’t need to tell PG what default DB to create with this env var. Ditto, I think, for the user.

Similarly, do we need them set in Ruby container’s env above? (`PGHOST` and `PGUSER`) Genuinely curious to know since I haven’t tried to do 2.0 on any projects yet.

frederikspang:
You're so welcome! I just felt inspired by digitalocean's hacktoberfest! :)

.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
...
+      - run: bundle install --path vendor/bundle
+
+      # Store bundle cache
+      - save_cache:
+          key: web-monitoring-db-{{ checksum "Gemfile.lock" }}
+          paths:
+            - vendor/bundle
+      # Database setup
+      - run: bundle exec rake db:create
+      - run: bundle exec rake db:schema:load
```
Mr0grog: I wonder if these two would be better combined with something like:

```yaml
- run:
    name: Set up database
    command: |
      bundle exec rake db:create
      bundle exec rake db:schema:load
```

Mr0grog:
Oh, you should also go ahead and add yourself to the [contributors section of the readme](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/README.md#contributors) :)

.circleci/config.yml
frederikspang: In CircleCI 2.0, it's docker based. Not having these lines will cause Rails to attempt running on a unix socket, and fail.

frederikspang:
Took a look at your review, I agree mostly! Except the Postgres ENV-vars, as I mentioned as a reply, will not set up login, as CircleCI2.0 is Docker-based, and thus won't allow connections to a unix-socket, we have to specify this.. I haven't found a way around it in any projects yet.. 

See also: [CircleCI on Postgres-config](https://circleci.com/docs/2.0/postgres-config/)

frederikspang:
I also just squashed the commits again.

.circleci/config.yml
Mr0grog: Well 💩! The [Ruby language guide](https://circleci.com/docs/2.0/language-ruby/) made me hope that wasn't the case.

Mr0grog:
Looks great; thanks so much for this, @frederikspang!

PR #149, opened 2017-10-08
Clean up migrations - db:setup seems to hang open after creating admi…

### LINKED ISSUE #69: Remove migrations related to `versionista_*`
…or just squash the migration history. There are a few migrations that use models for tables that were removed—their purpose was really to keep production data in the right shape. People keep trying to migrate an empty database from an initial state all the way to current, though, which does not work because of the removed models. (Rails style is to start a new database by loading the current schema rather than migrating from scratch.)

We should just squash/remove the no-longer-functional migrations so people can migrate across all of history.

katelovescode:
…n but the site looks like it's up - could use an additional tester

Fixes #69


Mr0grog:
Oooh, thanks so much for taking this on, @katelovescode!! I have to run right now, but will try and give this a review later tonight if I can.

katelovescode:
Thanks for the update on this; I'll give your ENV var suggestion a shot today and let you know how it goes.  Will also put in the fixes you requested and add myself to CONTRIBUTORS.md.  Stay tuned!

katelovescode:
Closing in favor of #153 which is connected to my remote repo; blame it on my use of bitbucket at work and rustiness with Github. ⚰️

PR #151, opened 2017-10-08
Require authentication for all API requests.

### LINKED ISSUE #122: Require authentication for everything.

danielballan:
Closes #122

As noted in #122, we should note in the API documentation
that authentication is required. @Mr0grog, di you have in mind
an overall statement about this at the top, a change to swagger.yml,
or both?

### DIFF
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -23,7 +23,7 @@ class Api::V0::ApiController < ApplicationController
   # This is different from Devise's authenticate_user! -- it does not redirect.
   def require_authentication!
     unless user_signed_in?
-      render_errors('You must be logged in to perform this action.', 401)
+      raise Api::AuthorizationError, 'You must be logged in to perform this action.'
     end
   end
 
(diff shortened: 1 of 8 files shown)


app/controllers/api/v0/api_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Api::V0::ApiController < ApplicationController
   include PagingConcern
-  before_action :require_authentication!, only: [:create]
+  before_action :require_authentication!
```
Mr0grog: While we’re here, I think it would be good to adjust the `require_authentication!` method to actually raise an `Api::AuthorizationError` (you’d need to create that in `lib/api/api.rb`) or `Api::DynamicError`. I should have written it that way from the beginning, sorry :\

danielballan: No problem! Done, covered with a test, force-pushed.

danielballan:
I'm not sure I've phrased e22f4b4 very well. I'd be glad to revise that if you have suggestions. Otherwise, merge if you are happy.

Mr0grog:
That looks good. I added some more detailed auth instructions and very light styling. Merging.

PR #153, opened 2017-10-10
Feature/cleanup db migrations

### LINKED ISSUE #69: Remove migrations related to `versionista_*`

katelovescode:
GRAH - so I messed up and deleted my local before waiting for it to be merged in.  I'm going to close the other PR but reference here -  #149 is the first attempt, please see conversation there.

MEANWHILE.  This has the update - I did *not* save the schema to this branch because no matter what I've done to the extensions to try and remove `uuid-ossp` and replace it with `pgcrypto`, it still generates the schema with this:
```
  # These are extensions that must be enabled in order to support this database
  enable_extension "plpgsql"
  enable_extension "pgcrypto"
  enable_extension "uuid-ossp"
```
and creation statements like this:
```
create_table "annotations", primary_key: "uuid", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
```
`uuid_generate_v4()` is the method from `uuid-ossp`.  As you can see in the PR, I tried to default to `pgcrypto`'s `gen_random_uuid()` but it doesn't seem to want to listen to me.  *Entirely* possible I missed the right documentation to get this figured out, would be very glad of feedback.  Postgres extensions are beyond my experience!

Also, fwiw, your recommendation about the ENV var worked, @Mr0grog 

Fixes #69


db/migrate/20170313192244_create_new_db_schema_with_generic_page_version_change_annotation.rb
```diff
@@ -1,10 +1,11 @@
...
-    # Postgres needs this for proper UUID support
-    enable_extension 'uuid-ossp'
+    # Since 9.4, PostgreSQL recommends using `pgcrypto`'s `gen_random_uuid`
+    # https://github.com/rails/rails/commit/b915b11cca558eb99b7c2621c4457491d4bdb43b
+    enable_extension 'pgcrypto'
 
-    create_table :pages, id: false do |t|
-      t.primary_key :uuid, :uuid
+    create_table :pages, id: false, default: -> { 'gen_random_uuid' } do |t|
+      t.primary_key :uuid, gen_random_uuid
```
Mr0grog: Hmmm, I don’t think you actually need to make these changes to the `default` option for `create_table` or the type of the primary key.

For me, `schema.rb` gets updated showing that `gen_random_uuid()` is being used, but only if the migrations don’t fail (in which case I think it’s just that `schema.rb` never gets written out to disk). Since migration 20170517222310 is failing, you need to not run that, for example—I tried `rake db:migrate VERSION=20170508221326`.

I think the behavior might also depend on what version of PG you have. I’m running 9.6.

db/migrate/20170313192244_create_new_db_schema_with_generic_page_version_change_annotation.rb
```diff
@@ -1,10 +1,11 @@
 class CreateNewDbSchemaWithGenericPageVersionChangeAnnotation < ActiveRecord::Migration[5.0]
   def change
-    # Postgres needs this for proper UUID support
-    enable_extension 'uuid-ossp'
+    # Since 9.4, PostgreSQL recommends using `pgcrypto`'s `gen_random_uuid`
+    # https://github.com/rails/rails/commit/b915b11cca558eb99b7c2621c4457491d4bdb43b
+    enable_extension 'pgcrypto'
```
Mr0grog: Is it safe to *remove* `uuid-ossp` here? (I actually don’t know.) I’m worried we’ll be missing something if we do. My `schema.rb` winds up having the `uuid-ossp` line removed, too, which also seems different from what you’re experiencing :\

db/migrate/20170517222310_change_annotation_author_to_not_null.rb
```diff
@@ -2,6 +2,7 @@ class ChangeAnnotationAuthorToNotNull < ActiveRecord::Migration[5.1]
   def do_sql(*args)
     expression = ActiveRecord::Base.send :sanitize_sql, args
     ActiveRecord::Base.connection.exec_query(expression)
+    raise StandardError, 'You can’t run this migration without any users in your DB'
```
Mr0grog: This will always raise, which is definitely a problem. I was thinking in `up`, you might use `.first.try(:[], 'id')` to get the user ID, then raise the error if it’s `nil`. e.g:

```rb
def up
  default_author_id = do_sql(
    'SELECT id FROM users ORDER BY created_at ASC LIMIT 1'
  ).first.try(:[], 'id')

  if default_author_id.nil?
    raise StandardError, 'You can’t run this migration without any users in your DB'
  end
```

Mr0grog:
Also, sorry I wasn’t able to get back to reviewing this last night!

db/migrate/20170517222310_change_annotation_author_to_not_null.rb
Mr0grog: Hmm, maybe simpler and better to not use `try` and just separate the steps of getting the user and the user’s ID:

```rb
default_author = do_sql(
  'SELECT id FROM users ORDER BY created_at ASC LIMIT 1'
).first

if default_author.nil?
  raise StandardError, 'You can’t run this migration without any users in your DB'
end

do_sql(
  'UPDATE annotations SET author_id = ? WHERE author_id IS NULL',
  default_author['id']
)
```

katelovescode:
Thanks again for the detailed feedback; I will give this a look and make some adjustments ASAP

katelovescode:
```
$ ruby --version
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
$ rails --version
Rails 5.1.4
$ postgres --version
postgres (PostgreSQL) 10.0
```
Still can't get my schema to update, possibly because of the postgres version, I suppose.  I could downgrade and try again, but would you mind giving it a shot on your local setup?

katelovescode:
Oh, also, I was testing by doing:
```
bin/rails db:environment:set RAILS_ENV=development
rake db:drop
rake db:create
rake db:migrate
```
which throws the following error for me:

```
== 20170517222310 ChangeAnnotationAuthorToNotNull: migrating ==================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

undefined method `[]' for nil:NilClass
```

When I run `ALLOWED_ARCHIVE_HOSTS='https://edgi-versionista-archive.s3.amazonaws.com/' bundle exec rake db:setup` I get no errors but my schema doesn't update.

I do think this might be a case of environment differences, but I'm willing to debug locally further if you'd like me to!

README.md
```diff
@@ -123,6 +123,7 @@ If you are getting errors such as `FATAL: role "user" doesn't exist. Couldn't cr
     sudo -u postgres psql -c "CREATE USER \"web-monitoring-db_development\" WITH PASSWORD 'wmdb';"
     sudo -u postgres createdb -O web-monitoring-db_development web-monitoring-db_development -E utf-8
     sudo -u postgres psql -c 'CREATE EXTENSION IF NOT EXISTS "uuid-ossp";' web-monitoring-db_development
+    sudo -u postgres psql -c 'CREATE EXTENSION IF NOT EXISTS "pgcrypto";' web-monitoring-db_development
```
Mr0grog: Ooh! Glad you caught this. I had totally forgotten about it. 👍

db/migrate/20170517222310_change_annotation_author_to_not_null.rb
```diff
@@ -9,6 +9,10 @@ def up
       'SELECT id FROM users ORDER BY created_at ASC LIMIT 1'
     ).first['id']
 
+    if default_author.nil?
```
Mr0grog: This line itself raises an error—there’s no `default_author` variable. To do this, you still need to change where `default_author_id` is defined above, so you have:

```rb
default_author = do_sql('same query').first
```

…and where it is used below:

```rb
do_sql('some query', default_author['id'])
```

This is also the reason you were getting the `undefined method '[]' for nil:NilClass` error—`do_sql(...).first` was `nil` (no users in the DB), so you can’t call `['id']` on it. To work your way around this, you can:

```sh
rake db:create
rake db:migrate VERSION=20170508221326 # stop before the author not null migration
rails c
# In the Rails console, create a new user like so, then quit:
# User.create(email: 'someone@example.com', password: 'password', confirmed_at: Time.now)
rake db:migrate
```

(You could also change this migration to *create* a user if one is not present.)

katelovescode: Cool.  Sorry for so much back-and-forth on this; context switching between my work and this means I'm missing some easy stuff.  I'll fix this bug and run again.  Also, thanks for the explanation on db:setup; I'm used to working in a local env where we have a make task that re-migrates every time, I forgot that wasn't default behavior. :)  Thanks for your patience and your help, will fix and push again tonight hopefully!

katelovescode:
VICTORY.

I seemed to have to do this every time right before I ran a db:drop to try a fresh setup:
`bin/rails db:environment:set RAILS_ENV=development`

Then I'd do:
```
bundle exec rake db:drop
ALLOWED_ARCHIVE_HOSTS='https://edgi-versionista-archive.s3.amazonaws.com/' bundle exec rake db:setup
bundle exec rake db:migrate
```

All successes and the schema is correct.  Since you suggested adding a user if it's not found, I did so; I like the idea of creating a fresh author if you're setting up from scratch.

I tried duplicating the `.env.example` to `.env` but it didn't seem to pull in the exported ALLOWED_ARCHIVE_HOSTS constant.  I didn't see `dotenv` in the gemfile so I suspect you're implementing this in a way I'm not familiar with.

HTH.

Mr0grog:
Looks great! Gonna merge this in.

> I tried duplicating the .env.example to .env but it didn't seem to pull in the exported ALLOWED_ARCHIVE_HOSTS constant. I didn't see dotenv in the gemfile so I suspect you're implementing this in a way I'm not familiar with.

Ah, yeah, we don’t do anything with it—you should `source .env` before running. There was some earlier back-and-forth about dotenv and I’ve been meaning to get around to putting it in for a while now :\

katelovescode:
Make an issue out of it!  I can handle that for Hacktoberfest too!

katelovescode:
Also - thanks x100 for everything throughout this process.  I don't typically contribute to Open Source and you really made it an enjoyable experience. :)

Mr0grog:
Thanks! Added an issue for dotenv: #160

Issue #154, opened 2017-10-10
Investigate rails performance improvements


patcon:
Reticketed from https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/119#issuecomment-335305906

Just wanted to create a stub for persons who might go down this path and take a swing at improving performance.

@Mr0grog Am I reading you correctly that you don't think it's a low-hanging fruit situation, and so not worth simply diving in with `rack-mini-profiler` or anything?

Do you think we'd get value only from an expert, or could it benefit from the attention of someone interested in learning about rails performance? (I ask as I was thinking on issues that might be good for [Hacktoberfest](http://hacktoberfest.digitalocean.com/).)

Mr0grog:
> Am I reading you correctly that you don't think it's a low-hanging fruit situation, and so not worth simply diving in with rack-mini-profiler or anything?

No, it could totally be that! In fact, my primary suspicion is that it is either the default JSON serializer; so switching to Oj might be an easy solution, or something about how AR is deserializing from the DB (probably much harder to fix). A simple profiling run would at least tell us where the problem lies.

patcon:
ok rad! I'll add some rando resources that I found, and maybe we could tag this as a first-timer/hacktoberfest!

Some ideas for how this ticket could be acted upon (feel free to open individual issues for these):
- Set up the [`rack-mini-profiler`](https://github.com/MiniProfiler/rack-mini-profiler) for the project and [learn a bit about it](https://github.com/MiniProfiler/rack-mini-profiler#learn-more), then run through [a tutorial using our app](https://www.speedshop.co/2015/08/05/rack-mini-profiler-the-secret-weapon.html)
- Add [Rails performance tests](http://guides.rubyonrails.org/v3.2.13/performance_testing.html)
- Something else of your choosing -- we trust you!

weatherpattern:
Closing. If we need to revisit rail performance we will open a new issue.

PR #162, opened 2017-10-16
Big perf/memory improvements in pages controller

Mr0grog:
Generally addresses #154.

This started out with some simple profiling and the addition of `Oj`, a faster JSON (de)serializer. That had a nice speed boost of just under 2x, so then I went a little nuts to bring memory way down (4-5x) reduction) and speed way up (about 5x). I don’t know that I’d recommend this approach for general use, but the page-related queries are *really* hot and desperately need all the optimization they can get, so I felt it made sense here.

In the following comparisons, “base” is the current, standard, Rails-y approach, while “Oj” is simply adding Oj (and restructuring slightly to get better use of it) in [the first commit](https://github.com/edgi-govdata-archiving/web-monitoring-db/commit/ed179545f7cf0e56ac50bad5b2d3d2377278ddb4), and “nutty optimizations” is [the second commit](https://github.com/edgi-govdata-archiving/web-monitoring-db/commit/db8a454003410d84203028f81bfd6aeb9491a625), where I went off the deep end and ditched ActiveRecord models entirely. I also made times slightly less precise (seconds instead of milliseconds) because that gave us a small but meaningful boost. “Precise time” is with millisecond precision.

(Local, production-mode) Performance on `GET /api/v0/pages`

| Approach            | Allocated Bytes | Allocated Objects | Retained Bytes | Retained Objects | Avg. Time (ms) |
| ------------------- | --------------: | ----------------: | -------------: | ---------------: | -------------: |
| Base                | 1,746,940       | 25,773            | 41,260         | 98               | 30.0 |
| Oj                  | 997,007         | 12,449            | 41,218         | 98               | 25.3 |
| Nutty Optimizations | 466,209         | 5,999             | 40,598         | 98               | 11.0 |

(Local, production-mode) Performance on `GET /api/v0/pages?include_versions`

| Approach                           | Allocated Bytes | Allocated Objects | Retained Bytes | Retained Objects | Avg. Time (ms) |
| ---------------------------------- | --------------: | ----------------: | -------------: | ---------------: | -------------: |
| Base                               |      36,800,463 |           510,597 |        811,382 | 99 | 454.4 |
| Oj                                 |      17,740,168 |           199,862 |        811,340 | 99 | 273.4 |
| Nutty Optimizations                |       7,916,625 |            92,617 |        800,324 | 99 |  91.3 |
| Nutty Optimizations - precise time |       8,380,853 |           100,714 |        811,120 | 99 | 109.1 |

It’s important to note that this is performance on my relatively speedy, memory-filled, and unburdened local machine (though in production mode). Our servers (especially staging) are waaaaaay slower and lower memory; they regularly get into the 10s response range and often exhaust their memory on `?include_versions` queries, which is why this is important.

And here are some super-ugly flame charts to demonstrate how this plays out in terms of stack:

flame-charts

(Remember that the width of each of these represents the length of a single request, not absolute time. To compare absolute time, shrink each of them to ~50% of the width of the one above.)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,6 +21,9 @@ Metrics/ParameterLists:
   Exclude:
     - 'lib/versionista_service/*'
 
+Metrics/CyclomaticComplexity:
+  Max: 12
+
 Layout/AlignParameters:
   EnforcedStyle: with_fixed_indentation
 
--- a/Gemfile
+++ b/Gemfile
@@ -18,11 +18,12 @@ gem 'rack-cors', :require => 'rack/cors'
 gem 'resque'
 gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.3.0'
+gem 'oj', '~> 3.3'
 
 # See https://github.com/rails/execjs#readme for more supported runtimes
 # gem 'therubyracer', platforms: :ruby
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
-gem 'jbuilder', '~> 2.6'
+# gem 'jbuilder', '~> 2.6'
 # Use Redis adapter to run Action Cable in production
 # gem 'redis', '~> 3.0'
 # Use ActiveModel has_secure_password
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -73,9 +73,6 @@ GEM
     httparty (0.15.6)
       multi_xml (>= 0.5.2)
     i18n (0.8.6)
-    jbuilder (2.7.0)
-      activesupport (>= 4.2.0)
-      multi_json (>= 1.2)
     jmespath (1.3.1)
     json (2.1.0)
     jwt (1.5.6)
@@ -101,6 +98,7 @@ GEM
     nio4r (2.1.0)
     nokogiri (1.8.1)
       mini_portile2 (~> 2.3.0)
+    oj (3.3.8)
     orm_adapter (0.5.0)
     parallel (1.12.0)
     parser (2.4.0.0)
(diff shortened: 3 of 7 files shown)


danielballan:
Nice work! Needs a corresponding fix to the Python API to add the precise_times option and check that it copes with the new time format. (It should be fine.)

Mr0grog:
> Needs a corresponding fix to the Python API to add the precise_times option

Hmmm, not sure if I’m misunderstanding what you wrote or if I didn’t write very clearly. There’s not a precise time option here; I just turned down the precision, period. (But included a benchmark for keeping it on.) Or is that an option on the Python side?

> and check that it copes with the new time format.

Let me know when you’ve done that :)

I think my biggest question here is whether the second commit (the “nutty optimizations”) seem OK to you. They are a little out there in terms of reasonability (one step shy of, and maybe in some ways slightly more complicated than, writing straight SQL queries) and definitely not perfectly generic. I think they were worthwhile for pages, but I intentionally didn’t use them for any of the other API endpoints because of that complexity (though hopefully they are written generically enough that we could apply them elsewhere if needed).

Mr0grog:
(Also, this branch is running on staging right now if you want to poke at it. It seems to be helping, but is still imperfect—though I think that’s a factor of staging being on hobby dev dynos, which are extra slow, low memory, and may randomly go to sleep.)

danielballan:
To echo my comments on Slack for the record here, I have some concerns about the long-term maintenance of code depending on the Rails expertise the project, but I appreciate that the "nutty" optimizations are isolated to one commit, and I'm confident we could wind them back if it comes to that.

I confirmed that the Python API still works with this change. Merging!

PR #166, opened 2017-10-16
Add next link to VersionsController#show

### LINKED ISSUE #164: Single version response should have `next` in `links` object
The `links` object for a single version `GET /api/v0/versions/{version_id}` or `GET /api/v0/pages/{page_id}/versions/{version_id}` has a `previous` property indicating the URL for the previous version. It should also have a `next` indicating the next version.

This should mainly be adding a property to the object in the [`versions_controller`](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/controllers/api/v0/versions_controller.rb#L16-L19) and implementing a `next` method analogous to the [`previous` method on the `Version` model](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/models/version.rb#L13-L15).

tylerferraro:
This closes #164

@edgi-govdata-archiving/lieutenants

### DIFF
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -15,7 +15,8 @@ class Api::V0::VersionsController < Api::V0::ApiController
     render json: {
       links: {
         page: api_v0_page_url(@version.page),
-        previous: @version.previous && api_v0_page_version_url(@version.page, @version.previous)
+        previous: @version.previous && api_v0_page_version_url(@version.page, @version.previous),
+        next: @version.next && api_v0_page_version_url(@version.page, @version.next)
       },
       data: @version
     }
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -15,8 +15,8 @@ class Api::V0::VersionsController < Api::V0::ApiController
     render json: {
       links: {
         page: api_v0_page_url(@version.page),
-        previous: @version.previous && api_v0_page_version_url(@version.page, @version.previous),
-        next: @version.next && api_v0_page_version_url(@version.page, @version.next)
+        previous: @version.previous && api_v0_version_url(@version.previous),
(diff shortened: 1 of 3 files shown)


app/controllers/api/v0/versions_controller.rb
```diff
@@ -15,7 +15,8 @@ def show
     render json: {
       links: {
         page: api_v0_page_url(@version.page),
-        previous: @version.previous && api_v0_page_version_url(@version.page, @version.previous)
+        previous: @version.previous && api_v0_page_version_url(@version.page, @version.previous),
+        next: @version.next && api_v0_page_version_url(@version.page, @version.next)
```
Mr0grog: This works great, but would you mind changing it to use `api_v0_version_url()` instead of `api_v0_page_version_url()` for both of these? That simpler route didn’t exist when we originally made this.

test/models/version_test.rb
```diff
@@ -10,4 +10,9 @@ class VersionTest < ActiveSupport::TestCase
     change = versions(:page2_v2).change_from_previous
     assert_not_nil change
   end
+
+  test 'next should get the next version' do
+    nxt = versions(:page1_v1).next
```
Mr0grog: This is a really minor nit, but we try and avoid most abbreviations. I know `next` is a keyword in Ruby, but how about `url` or `next_url` or something similar?

tylerferraro: I changed it to `next_version` since the object is a `Version` object. That way it's not confused for an actual url path.

Mr0grog: Oh, ha, you’re right! I was stuck thinking about the data in the response :P

Mr0grog:
Did you want to add yourself to the contributors list here, or should I go ahead and merge this?

tylerferraro:
I am not interested in being added to the contributor list. Go ahead and merge, thanks!

Mr0grog:
OK. Thanks again.

PR #167, opened 2017-10-16
Feature/change id from

### LINKED ISSUE #157: Allow change IDs using only the “from” version
Right now, you can specify changes in the API via their “from” and “to” versions like so: `{from_id}..{to_id}`

You can also specify just the “to” version, in which case it’s treated as a change between the “to” version and version immediately prior to it: `..{to_id}`.

It would be nice to also be able to do the opposite and specify only the “from” version, implying that the “to” is the next version: `{from_id}..`.

This requires:

- Adding `next` and `change_to_next` methods to [`app/models/version.rb`](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/models/version.rb) to mirror the existing `previous` and `change_from_previous` methods.
- Using the above methods to support the new syntax in [`find_by_api_id` in `app/models/change.rb`](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/models/change.rb#L29-L44)

tylerferraro:
This change is built on top of #166 and closes #157

@edgi-govdata-archiving/lieutenants

### DIFF
--- a/app/models/change.rb
+++ b/app/models/change.rb
@@ -31,8 +31,12 @@ class Change < ApplicationRecord
 
     if api_id.include?('..')
       from_id, to_id = api_id.split('..')
-      if from_id.present?
+      if from_id.present? && to_id.present?
         Change.between(from: from_id, to: to_id)
+      elsif from_id.present?
+        Version.find(from_id).change_from_next ||
+          (raise ActiveRecord::RecordNotFound, "There is no version following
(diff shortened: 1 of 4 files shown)


tylerferraro:
Updated the route constraint to allow the new url style.

Mr0grog:
Great, thanks so much for this!

Issue #150, opened 2017-10-08
Add `meta` section to list responses with count of results


Mr0grog:
List-style responses (e.g. `/api/v0/pages`) that are paginated should probably have a [`meta` section](http://jsonapi.org/format/#document-meta) that includes a count of all the results. This would have made writing an API-based script for edgi-govdata-archiving/web-monitoring#74 easier, for example.

shishir127:
@Mr0grog I can help with this but I need more information to understand this. I tried to go through the linked issue but it's too much info to process for a noob like me.

1. The client using this API can just parse the JSON and find out the size of the list e.g. we can invoke `response_body['data'].size` on the parsed JSON reponse of the pages API. Is the `meta` section with a `count` really required here?
2. Can you give me an exhaustive list of which APIs need the `meta` section (assuming it's needed in other APIs)?

Mr0grog:
> The client using this API can just parse the JSON and find out the size of the list

Actually, they can’t! All these APIs are paginated, so a single HTTP response may not contain the entire list of results (you can move through the results using the `chunk` parameter). In the response, the `links` object has important pagination-related info:

- `first` - URL of first chunk of this result set
- `last` - URL of the last chunk of this result set
- `previous` - URL of the previous chunk of this result set (i.e. if you are on chunk 5, this is the URL for chunk 4)
- `next` - URL of the next chunk of this result set (i.e. if you are on chunk 5, this is the URL for chunk 6)

So `meta.count` would be the total number of results across all chunks. e.g. if the `chunk_size` is 100 and there are 3 chunks, there could be anywhere between 201 and 300 results, but you’d only get 100 if you just counted the number of items in `data`.

This is needed on every API that is a list:

- `/api/v0/pages`
- `/api/v0/pages/{page_id}/versions`
- `/api/v0/pages/{page_id}/changes`
- `/api/v0/pages/{page_id}/changes/{change_id}/annotations
- `/api/v0/versions`

Mr0grog:
Fixed in a3a1950455710ec3d854c36cb7a974264b73aa5c

PR #169, opened 2017-10-19
Meta section with count of results

shishir127:
This is for #150

@Mr0grog Please review and let me know if everything is correct and if this patch complies with code styling/convention.

### DIFF
--- a/test/controllers/api/v0/annotations_controller_test.rb
+++ b/test/controllers/api/v0/annotations_controller_test.rb
@@ -202,7 +202,7 @@ class Api::V0::AnnotationsControllerTest < ActionDispatch::IntegrationTest
     )
     assert_response :success
 
-    get(api_v0_page_change_annotations_path(page, "..#{change_id}", params: {chunk_size: 1}))
+    get(api_v0_page_change_annotations_path(page, "..#{change_id}", params: { chunk_size: 1 }))
 
     assert_response :success
     body_json = JSON.parse @response.body
--- a/test/controllers/api/v0/annotations_controller_test.rb
+++ b/test/controllers/api/v0/annotations_controller_test.rb
@@ -202,6 +202,7 @@ class Api::V0::AnnotationsControllerTest < ActionDispatch::IntegrationTest
     assert_equal(
       Change.find_by_api_id("..#{change_id}").annotations.count,
       body_json['metadata']['total_results'],
-      'Should contain the count of total results across all pages')
+      'Should contain the count of total results across all pages'
+    )
   end
 end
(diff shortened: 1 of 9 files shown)


test/controllers/api/v0/annotations_controller_test.rb
```diff
@@ -166,4 +166,43 @@ class Api::V0::AnnotationsControllerTest < ActionDispatch::IntegrationTest
...
+    assert_response :success
+
+    get(api_v0_page_change_annotations_path(page, "..#{change_id}"))
+
+    assert_response :success
+    body_json = JSON.parse @response.body
+
+    assert_equal(
+      Change.find_by_api_id("..#{change_id}").annotations.count,
+      body_json['metadata']['total_results'],
```
Mr0grog: This isn’t really adequate—if there’s only one annotation, all the annotations will always be in the response. It might be better to post two annotations, then use the `chunk_size` parameter in the request to limit it to only one result (i.e. `get(api_v0_page_change_annotations_path(page, "..#{change_id"}, params: {chunk_size: 1}))`).

Most of the other tests should probably do the same with `chunk_size`.

Mr0grog:
Hey @shishir127, are you come back around and do another pass at this?

shishir127:
Yup, will do. Sorry was out for a vacation. Just came back this week.

Mr0grog:
No worries, just checking in. 👍

shishir127:
@Mr0grog everything look fine now? made the changes you wanted.

test/controllers/api/v0/versions_controller_test.rb
```diff
@@ -175,4 +176,18 @@ class Api::V0::VersionsControllerTest < ActionDispatch::IntegrationTest
       'The returned version did not have a matching ID'
     )
   end
+
+  test 'meta property' do
```
Mr0grog: Same note about descriptive naming as for the test for annotations above.

Also, this has the same issue as the annotation test used to—we should make sure to pick a `chunk_size` value that ensures we would have multiple pages of results.

shishir127:
@Mr0grog I've made the changes. Let me know if it looks good.

Mr0grog:
Looks perfect. Thanks so much, @shishir127! Do you want to add yourself to the [contributors list in the README](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/README.md#contributors)? If so, go ahead and add another commit with that. Otherwise, I’ll go ahead and merge this now.

shishir127:
I've updated the contributors list in the readme. You can go ahead an merge this @Mr0grog . Thanks for your help.

PR #163, opened 2017-10-16
Add the dotenv gem

kholbekj:
As per #160 :)

PR #170, opened 2017-10-22
add dotenv gem

### LINKED ISSUE #160: Add dotenv gem
As a follow-on to discussions in #17, we really should add dotenv to make sure `.env` files get read in automatically instead of requiring people to explicitly `source` them before running the app or rake tasks. Hopefully this will eliminate a lot of simple mistakes.

- Add dotenv gem
- Add info to readme about copying `.env.example` so there’s something for dotenv to use

gregmolnar:
closes #160


Mr0grog:
Hi @gregmolnar, thanks for the effort! However, this is already underway on #163. Please make sure you check the list of PRs (or check the issue for attached PRs) before working on an issue.

PR #172, opened 2017-11-04
Create sorting concern for index endpoints

### LINKED ISSUE #165: Add a parameter to control ordering on list requests
We should have a relatively standard way to control ordering on requests that return a list of pages/versions/changes/annotations. This could be as simple as ascending/descending or as complex as a list of fields and sort directions, e.g. `capture_time:desc,title:asc` or something.

This needs a clearer design proposal before implementation.

tylerferraro:
Closes #165

### DIFF
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -1,8 +1,10 @@
 class Api::V0::PagesController < Api::V0::ApiController
   def index
     query = page_collection
+    sorting = sortation(params[:sort])
     paging = pagination(query)
     pages = query.limit(paging[:chunk_size]).offset(paging[:offset])
+    pages = pages.order(sorting) unless sorting.nil?
 
     # In order to handle pagination when querying across pages and versions
     # together, we do two separate queries:
--- a/app/controllers/concerns/sorting_concern.rb
+++ b/app/controllers/concerns/sorting_concern.rb
@@ -1,0 +1,30 @@
+module SortingConcern
+  extend ActiveSupport::Concern
+
+  protected
+
+  # Sorting params example: capture_time:desc,title:asc
+  def sortation(sort_params)
+    return if sort_params.nil?
+
+    sort_params.split(',')
+               .map { |sort| ordering(sort) }
+               .join(', ')
+    end
+  end
+
+  private
+
+  def ordering(sort)
+    key, direction = sort.split(':')
+
+    "#{key} #{sanitize_direction(direction)}"
+  end
+
+  def sanitize_direction(direction)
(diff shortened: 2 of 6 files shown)


Mr0grog:
Sorry for being so slow to get to this, @tylerferraro; I’ll try to take a closer look tonight. I’m guessing the failures might be a conflict with the sorting in the [`lightweight_query` functionality](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/controllers/api/v0/pages_controller.rb#L129-L135).

app/controllers/api/v0/pages_controller.rb
```diff
@@ -1,8 +1,10 @@
 class Api::V0::PagesController < Api::V0::ApiController
   def index
     query = page_collection
+    sorting = sortation(params[:sort])
     paging = pagination(query)
     pages = query.limit(paging[:chunk_size]).offset(paging[:offset])
+    pages = pages.order(sorting) unless sorting.nil?
```
Mr0grog: I think the sorting may need to be added to `query`, not `pages` here, since the workaround for an ActiveRecord bug immediately below this code uses `query` instead of `pages` to build the final SQL query.

This also definitely needs tests, since that would reveal that the built-in sorting for the [`lightweight_query` functionality](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/controllers/api/v0/pages_controller.rb#L129-L135) for this controller will conflict with and override any sorting provided here. I think you’ll need to fix that, too.

app/controllers/concerns/sorting_concern.rb
```diff
@@ -0,0 +1,30 @@
...
+  protected
+
+  # Sorting params example: capture_time:desc,title:asc
+  def sortation(sort_params)
+    return if sort_params.nil?
+
+    sort_params.split(',')
+               .map { |sort| ordering(sort) }
+               .join(', ')
+    end
```
Mr0grog: Looks like all the test failures are due to this spurious `end` statement.

app/controllers/concerns/sorting_concern.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  private
+
+  def ordering(sort)
+    key, direction = sort.split(':')
+
+    "#{key} #{sanitize_direction(direction)}"
```
Mr0grog: This does not seem safe. [ActiveRecord does not sanitize single string arguments to `order()`](https://github.com/rails/rails/blob/v5.1.4/activerecord/lib/active_record/sanitization.rb#L64-L70). Maybe better to produce a hash of `key => direction` here so we get the benefit of AR’s built-in sanitization and safety? Or quote the key?

PR #176, opened 2017-11-29
Remove uuid+site, add site-only index on pages

Mr0grog:
I have no idea what possessed me to make a multi-column index like this, since we never do a query on `UUID` + `site`. Instead, just index `site`. Tried this out on staging and it seems to have made a reliable performance improvement.

(Also totally forgot to push this! It’s been sitting around for a while.)


lightandluck:
FWIW, makes sense and looks good to me. I was able to run the migration and saw my indexes change, but have no idea how to test if it made things faster for me locally.

Issue #182, opened 2017-12-06
Add support for link diff


Mr0grog:
DB needs to support the new diff of outgoing links in a page: https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/131

Issue #181, opened 2017-12-06
Diff `data` field should be exactly what the diff server sent


Mr0grog:
At current, DB nests the actual diff server’s response in [another object with more metadata](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/89#issuecomment-329353425). However, we have since [added diff server version data](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/89#issuecomment-329353425) into the diff server’s response and are considering adding more metadata at that level with [a count of changes](https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/130).

This has lead to some extra-deep structures for the returned data:

```js
{
  "data": {
    "page_id": "9420d91c-2fd8-411a-a756-5bf976574d10",
    "from_version_id": "4b089be5-c6ef-44ce-9a63-fd52443bea3b",
    "to_version_id": "6b600c39-29cd-4fad-9a3a-5edc504f61d8",
    "diff_service": "html_text",
    "content": {
      "diff": { /* actual diff here */ },
      "version": "0+untagged.292.g7062697"
    }
  }
}
```

It also means the structure of a diff returned from DB is very different from that returned from the diff server. Another approach we have talked about is whether the UI could or should talk directly to the diff server.

I think it might be an improvement to merge the DB’s added metadata with the content from the diff server. That at least makes things slightly simpler:

```js
{
  "data": {
    "page_id": "9420d91c-2fd8-411a-a756-5bf976574d10",
    "from_version_id": "4b089be5-c6ef-44ce-9a63-fd52443bea3b",
    "to_version_id": "6b600c39-29cd-4fad-9a3a-5edc504f61d8",
    "diff_service": "html_text",
    "diff": { /* actual diff here */ },
    "version": "0+untagged.292.g7062697"
  }
}
```

We could also restructure this to keep the diff content as much as-is as possible:

```js
{
  "links": {
    "page": "https://api.monitoring.envirodatagov.org/api/v0/pages/9420d91c-2fd8-411a-a756-5bf976574d10",
    "from_version": "https://api.monitoring.envirodatagov.org/api/v0/versions/4b089be5-c6ef-44ce-9a63-fd52443bea3b",
    "to_version": "https://api.monitoring.envirodatagov.org/api/v0/versions/6b600c39-29cd-4fad-9a3a-5edc504f61d8"
  },
  "meta": {
    "diff_service": "html_text"
  },
  "data": {
[...]

PR #184, opened 2017-12-09
Add a “default” differ type using `DIFFER_DEFAULT`

### LINKED ISSUE #180: Make it possible to forward any diff to a given diff server
Right now, we have to explicitly configure each and every diff type, which means that, whenever the `processing` project adds a new diff, we have to at least update the configuration of DB servers (if not also change code), which is kind of ridiculous. It also leads to configuration like:

```sh
DIFFER_LENGTH=http://localhost:8888/length
DIFFER_IDENTICAL_BYTES=http://localhost:8888/identical_bytes
DIFFER_PAGEFREEZER=http://localhost:8888/pagefreezer
DIFFER_HTML_SOURCE=http://localhost:8888/html_source_diff
DIFFER_HTML_TEXT=http://localhost:8888/html_text_diff
DIFFER_SIDE_BY_SIDE_TEXT=http://localhost:8888/side_by_side_text
DIFFER_HTML_VISUAL=http://localhost:8888/html_visual_diff
DIFFER_LINKS=http://localhost:8888/links_diff
```

Which feels pretty darn redundant.

We should support a single env var (`DIFF_SERVER_URL`?) that is used for any unrecognized diff name in:

```
GET /api/v0/pages/{page id}/changes/{change id}/diff/{diff name}
```

Mr0grog:
The value of the `DIFFER_DEFAULT` environment variable will be used as the URL for a new SimpleDiff instance if no diff has been registered for a requested diff type. You can still keep explicitly registered differs via any environment variables named `DIFFER_*`; they’ll be matched before falling back to creating a default differ. This way, we can keep current configurations in place until the UI has updated to request the correct diff names (DB’s names don't quite match Processing’s names; see https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/133).

This also tightens up error handling of responses from the differ (since we’ll see them more often now). We should get on edgi-govdata-archiving/web-monitoring-processing#121 to make this better still.

Fixes #180 and #182.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,13 +39,7 @@ ALLOWED_ARCHIVE_HOSTS='https://edgi-versionista-archive.s3.amazonaws.com/ https:
 # If you are developing this web-monitoring-db Rails app + wm-differ-differ locally,
 # then settings below are the default settings. If you are running against a
 # production deployment, change 'localhost:8888' to the production URL.
-DIFFER_LENGTH=http://localhost:8888/length
-DIFFER_IDENTICAL_BYTES=http://localhost:8888/identical_bytes
-DIFFER_PAGEFREEZER=http://localhost:8888/pagefreezer
-DIFFER_HTML_SOURCE=http://localhost:8888/html_source_diff
-DIFFER_HTML_TEXT=http://localhost:8888/html_text_diff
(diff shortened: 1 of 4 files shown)


Mr0grog:
@danielballan if you don’t have any strong feelings on this right now, I’m going to go ahead and merge shortly. Can clean it up more when I come back around to do #181.

PR #186, opened 2017-12-12
Make diff endpoint's `data` the raw diff data

### LINKED ISSUE #181: Diff `data` field should be exactly what the diff server sent

Mr0grog:
The idea here is that all the metadata and extra info the DB server provides should be in the standard `links` and `meta` sections of the response, while `data` should just be exactly what you'd get if you called the diff server directly. Makes a lot more sense in hindsight.

If we ever do decide to expose the differ publicly or if we write other tools that access the differ internally, this helps ensure the interface is standard and code can be shared. It also hopefully forces us to put things more in the right place.

Fixes #181.

**NOTE: Do not merge yet. This will be a breaking change for the UI, so it needs updates to be deployed first. There are a lot of things in flight there, so I think this will be easier if we hold off just a bit on merging.**

### DIFF
--- a/app/controllers/api/v0/diff_controller.rb
+++ b/app/controllers/api/v0/diff_controller.rb
@@ -3,13 +3,15 @@ class Api::V0::DiffController < Api::V0::ApiController
     ensure_diffable
 
     render json: {
-      data: {
-        page_id: change.version.page.uuid,
-        from_version_id: change.from_version.uuid,
-        to_version_id: change.version.uuid,
-        diff_service: params[:type],
-        content: raw_diff
-      }
+      links: {
+        page: api_v0_page_url(change.version.page),
+        from_version: api_v0_version_url(change.from_version),
+        to_version: api_v0_version_url(change.version)
+      },
+      meta: {
+        requested_type: params[:type]
+      },
+      data: raw_diff
     }
   end
 
--- a/test/controllers/api/v0/diff_controller_test.rb
+++ b/test/controllers/api/v0/diff_controller_test.rb
@@ -3,6 +3,15 @@ require 'minitest/mock'
 
 class Api::V0::DiffControllerTest < ActionDispatch::IntegrationTest
   include Devise::Test::IntegrationHelpers
+
+  setup do
+    @original_default_differ = Differ.for_type(nil)
+  end
+
+  teardown do
+    Differ.register(nil, @original_default_differ)
+  end
+
   test 'can diff two versions' do
     sign_in users(:alice)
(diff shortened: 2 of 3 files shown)


Mr0grog:
Now that edgi-govdata-archiving/web-monitoring-ui#161, this is safe to merge.

PR #190, opened 2018-01-05
Log import errors and, separately, provide more info in dev mode.

danielballan:


app/jobs/import_versions_job.rb
```diff
@@ -9,8 +9,13 @@ def perform(import)
...
       import_raw_data(@import.load_data)
-    rescue StandardError => _
-      @import.processing_errors << 'Unknown error occurred'
+    rescue StandardError => error
+      @import.processing_errors << if Rails.env.development?
+                                     "Row #{row}: #{error.message}"
+                                   else
+                                     "Row #{row}: Unknown error occurred"
+                                   end
+      Rails.logger.error "Row #{row}: #{error.message}"
```
Mr0grog: Uh, there’s no row in this context to log anything about. This only gets errors generated by `each_json_line` (which splits up the import data into JSON objects). Errors that happen processing a given line/row are already handled in `import_raw_data`.

So I think:

1. The data you’re stuffing in `@import.processing_errors` here should be just either `error.message` or `Unknown error occurred`.
2. Likewise the `logger` call here needs to just be `error.message`
3. Both those logger calls should probably include the import ID and/or the import’s file URL, otherwise it’ll be nearly impossible to match up the bad data to the problem.

app/jobs/import_versions_job.rb
```diff
@@ -24,20 +29,25 @@ def import_raw_data(raw_data)
...
         @import.processing_errors << "Row #{row}: #{error.message}"
+        Rails.logger.error "Row #{row}: #{error.message}"
       rescue ActiveModel::ValidationError => error
         messages = error.model.errors.full_messages.join(', ')
         @import.processing_errors << "Row #{row}: #{messages}"
+        Rails.logger.error "Row #{row}: #{messages}"
       rescue ActiveRecord::RecordInvalid => error
         messages = error.model.errors.full_messages.join(', ')
         @import.processing_errors << "Row #{row}: #{messages}"
+        Rails.logger.error "Row #{row}: #{messages}"
```
Mr0grog: So with these errors, I very explicitly chose *not* to log (totally agree I should have for the unknown type of error), because they represent user input problems that a maintainer of the system should be bothered with or worry about. They’re only really useful as feedback to the user who imported data because they concern errors that user (*not* the app authors/maintainers) need to address.

app/jobs/import_versions_job.rb
```diff
@@ -24,20 +29,25 @@ def import_raw_data(raw_data)
...
       rescue StandardError => error
         @import.processing_errors << if Rails.env.development?
                                        "Row #{row}: #{error.message}"
                                      else
                                        "Row #{row}: Unknown error occurred"
                                      end
+        Rails.logger.error "Row #{row}: #{error.message}"
```
Mr0grog: See note above about including import ID or the import file’s URL.

app/jobs/import_versions_job.rb
Mr0grog: Actually, to this end, we could probably rescue `JSON::ParserError` here (emanating from line 145) and say something more meaningful about it than “unknown error.”

app/jobs/import_versions_job.rb
Mr0grog: (If following via e-mail, I missed a critical “not” in “a maintainer of the system should **not** be bothered.” Edited the comment on GH.)

danielballan:
Haha, it was late and I was in a gonzo "LOG ALL THE THINGS" frame of mind. I'll clean this up....

app/jobs/import_versions_job.rb
```diff
@@ -23,31 +23,27 @@ def perform(import)
     end
   end
 
-  def import_raw_data(raw_data)
+  def import_raw_data(raw_data, import_id)
```
Mr0grog: You could probably just use `@import.id` for the message instead of passing it here. That’s what we already do in each of the row errors below.

app/jobs/import_versions_job.rb
```diff
@@ -38,6 +43,7 @@ def import_raw_data(raw_data)
                                      else
                                        "Row #{row}: Unknown error occurred"
                                      end
+        Rails.logger.error "Import {@import} Row #{row}: #{error.message}"
```
Mr0grog: Missing `#` before `{@import}`. Also, it should be `#{@import.id}`, otherwise you’ll get the actual underlying Ruby object identifier, e.g. `#<Import:0x007f966df84ea8>`.

danielballan: Fixed, force-pushed.

Mr0grog:
Looks great to me!

Mr0grog:
Gonna go ahead and merge this shortly.

danielballan:
Fine by me!

PR #195, opened 2018-01-10
Add server-side caching of diffs

Mr0grog:
Builds on #194 to add server-side caching using a Redis store. This makes diffing giant half-megabyte and larger HTML pages like https://monitoring.envirodatagov.org/page/3ce1df73-126a-446c-b17e-2933fa42596b/6980cb93-1fbe-43a8-a29e-2ad8ac4f5ab7..10a83eb8-f19d-44d0-8d22-ac2efd6d4f40 a little more usable.

Use `REDIS_CACHE_URL` to set the Redis server to use for caching.


Mr0grog:
Clean Redis server for caching is already set up in EC2, awaiting usage when this merges.

PR #206, opened 2018-01-22
WIP: Multiple sites and agencies per page

### LINKED ISSUE #24: Handle pages that are tracked by multiple “sites”
`site` is currently a property of the `Page` model, but we already know there are a couple pages that are tracked by multiple Versionista sites. So far, we have stored these as separate Page records here, too, but as we are moving towards integrating multiple sources (#15), we need to handle this better. It’s been mentioned before that we should think of sites as something more akin to tags.

A few ideas:

1. Sites as another model with an N:M relationship to pages
2. Sites as a JSON array belonging to a page
3. Have a more generic tagging system and have tags like `site:[site name]` that are applied to pages

This also brings up the question: is `agency` similar? I’m thinking it’s fine as it is now—a direct property of a `Page`—but maybe worth thinking about in this context.

@danielballan Any thoughts on this? I’m thinking either 1 or 3 is best here.

Mr0grog:
**Work in progress; do not merge!**

I got three quarters of the way through writing a tags implementation before feedback from @trinberg suggested that tags might not actually be the best solution for modeling sites and agencies. Instead, I went back to the drawing board and just added a more straightforward set of two new models (Site and Agency) than each have an N:M relationship with Page.

Agency may as well be a simple tag since it only has a name, but this affords us room to add more metadata later. Site takes advantage of this and adds a concept of `versionista_id` (which we’ll hopefully remove in the future) that helps us handle sites being renamed in Versionista.

Still to do:

- [x] Add Agency and Site models
- [x] Add migration to move existing agency/site data to new models
- [x] Support new sites/agencies in imports
- [ ] Support multiple sites/agencies per record in imports
- [x] Support new sites/agencies in page results
- [ ] Support conditional queries on sites/agencies (e.g. `/api/v0/pages?agency=EPA`)

Fixes #24.

### DIFF
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -26,30 +26,32 @@ class Api::V0::PagesController < Api::V0::ApiController
           .where(uuid: page_ids)
           .includes(:versions)
           .order('versions.capture_time DESC')
-        lightweight_query(results, :versions)
+        lightweight_query(results)
       elsif should_include_latest
-        # [pages.includes(:latest), :latest]
-        page_ids = pages.pluck(:uuid, :updated_at).collect {|data| data[0]}
-        results = Page
-          .where(uuid: page_ids)
-          .includes(:latest)
-        lightweight_query(results, :latest)
+        pages.includes(:latest).as_json(include: [:agencies, :latest, :sites])
       else
         lightweight_query(pages)
       end
 
-    render json: Oj.dump({
-      links: paging[:links],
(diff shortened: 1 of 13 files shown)


Mr0grog:
I’m withdrawing this PR after the discussion and decision on a new direction here: https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/24#issuecomment-361043600

(Elements of this code will definitely be re-used for that, though.)

Mr0grog:
Superseded by #220.

PR #220, opened 2018-02-07
Replace agency → maintainers & site → tags

### LINKED ISSUE #24: Handle pages that are tracked by multiple “sites”

Mr0grog:
~**Work in progress; do not merge!**~

Fixes #24.

Replace the `agency` field on `Page` models with an N:M relationship to a new `Maintainer` model; replace the `site` field on `Page` models with an N:M relationship to a new `Tag` model.

- [x] Add `Maintainer`
- [x] Add `Tag`
- [x] Show page maintainers and tags in API
- [x] Make maintainers and tags queryable in pages API
- [x] Populate maintainers and tags during import
- [x] Copy existing site/agency data to new maintainer/tag models
- [x] Merge existing pages that are currently duplicated as a result of multiple sites or agencies
- [x] Add routes to list maintainers and tags:
    - [x] `GET /maintainers`
    - [x] `GET /maintainers/{uuid|name}`
    - [x] `PATCH|POST /maintainers/{uuid|name}`
    - [x] `GET /pages/{uuid}/maintainers`
    - [x] `POST /pages/{uuid}/maintainers`
    - [x] `GET /tags`
    - [x] `GET /tags/{uuid|name}`
    - [x] `PATCH|POST /tags/{uuid|name}`
    - [x] `GET /pages/{uuid}/tags`
    - [x] `POST /pages/{uuid}/tags`
- [x] Document new schemas, query params, and routes

Note: this PR intentionally leaves the old `site` and `agency` fields in place so as not to break existing clients. We’ll remove them later when we’re confident most clients are working in terms of maintainers/tags.

OK! This is most of the way done now, so I think it’s ready for feedback on whether this looks right. Some questions/things I’d really like feedback on here:

1. Do we need all the extra routes for listing/renaming/adding maintainers and tags as above?

2. Tag names are *not* case-sensitive (for querying, that is; the response shows the case they were originally set in), but maintainer names are. This made sense to me when I wrote it, but in hindsight, I’m not so sure.

3. Unlike versions, the page listing *always* includes associated maintainers and tags. I *think* that’s OK because we probably don’t expect there to be a lot of them, but who knows, that could be a terrible assumption.

4. Maintainers and tags are each exposed on page records as the full serialization of their object, e.g:

    ```js
    {
      /* page fields here */
      "maintainers": [
        {
          "uuid": "191da223-428f-44c7-9411-ef102c87b465",
          "name": "NASA",
          "parent_uuid": null,
          "created_at": "2018-02-07T06:22:43Z",
          "updated_at": "2018-02-07T06:22:43Z"
        }
      ],
      "tags": [
        {
          "uuid": "2ab826bb-99de-496d-bf5c-4f9e669c1522",
          "name": "site:NASA - science.nasa.gov - Earth Science",
          "created_at": "2018-02-07T06:22:43Z",
          "updated_at": "2018-02-07T06:22:43Z"
        }
      ]
    }
    ```

    Is that right? Should we just expose UUIDs or names? (e.g. `{maintainers: ["191da223-428f-44c7-9411-ef102c87b465"]}` or `{maintainers: ["NASA"]}`) (Is the right answer here different for maintainers and tags?)

    We could also expose a simplified list of names or UUIDs when attached to a page, but expose the full object when requesting `GET /pages/{uuid}/maintainers` / `GET /pages/{uuid}/tags`.

5. Related to (4), the association between pages and tags has one extra piece of metadata: the time it was created. Should tag records on pages be a wholly separate thing like:

    ```js
    {
      /* page fields here */
      "tags": [
        {
          "uuid": "2ab826bb-99de-496d-bf5c-4f9e669c1522",
          "name": "site:NASA - science.nasa.gov - Earth Science",
          "tagged_at": "2018-02-07T07:22:43Z"
        }
      ]
    }
    ```

    Note the `tagged_at` instead of `created_at`, implying this is the time the tag was added *to this page,* not the time the tag was first created.

6. Filtering: these are the first 1:N relation we have that’s queryable by some unique identifier (versions we only support filtering in ways that are non-unique). I’ve leaned on the standard way Rails does arrays in the querystring, so if you’re filtering by tags, it looks like:

    ```
    ?tags[]=tag1&tags[]=tag2

    # Basically the same for filtering by a single tag
    ?tags[]=tag1

    # Same pattern for maintainers
    ?maintainers[]=NASA&maintainers[]=DOE
    ```

    These lists are also OR (union) lists. So the above query gets all pages tagged with `tag1` and all pages tagged with `tag2`. It does *not* get you only the pages that are tagged with *both.*

    Is that right? Should we flip it to be AND (more complicated to implement, but I think doable)? Should we have some crazy complicated thing where you can do both, e.g:

    ```
    # Pages with: (tag1 AND tag2) OR (tag3 AND tag4)
    ?tags[]=tag1,tag2&tags[]=tag3,tag4
    ```

    My gut tells me this is getting overcomplicated for now.

7. Maintainers and tags both use UUIDs. At this point, I feel like the few places I used numeric IDs make things confusing, so I decided not to do that here, even though they don’t remotely have the constraints around quantity of values or ability to create an ID without already knowing what’s in the DB that pages and versions do.

/cc @danielballan

### DIFF
--- a/db/migrate/20180212043742_dedupe_pages_with_multiple_agencies_or_sites.rb
+++ b/db/migrate/20180212043742_dedupe_pages_with_multiple_agencies_or_sites.rb
@@ -1,0 +1,45 @@
+class DedupePagesWithMultipleAgenciesOrSites < ActiveRecord::Migration[5.1]
+  # Find all pages sharing the same URL and de-duplicate them by copying the
+  # maintainers, tags, and versions of all subsequent pages onto the one that
+  # was created first.
+  def up
+    total_deletions = 0
+
+    Page.group(:url).having('count(*) > 1').count.each_key do |url|
+      say("Deduplicating pages for '#{url}'")
+
+      canonical_page = nil
+      deletable = []
+      pages = Page.where(url: url)
+        .eager_load(:maintainers, :tags)
+        .order(created_at: :asc)
+
+      pages.each do |page|
+        if canonical_page.nil?
+          canonical_page = page
+        else
+          page.maintainers.each {|item| canonical_page.add_maintainer(item)}
+          page.maintainers.clear
+
+          page.tags.each {|item| canonical_page.add_tag(item)}
+          page.tags.clear
+
+          page.versions.update_all(page_uuid: canonical_page.uuid)
+          deletable << page.uuid
+        end
+      end
+
+      Page.where(uuid: deletable).delete_all
+      say("Deleted #{deletable.length} pages.", true)
+
+      total_deletions += deletable.length
+    end
+
+    say("Deleted #{total_deletions} pages in total.")
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration,
+      'This migration removed information and cannot be reversed.'
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,6 +23,8 @@ Metrics/ParameterLists:
 
 Metrics/CyclomaticComplexity:
   Max: 12
+  Exclude:
+    - 'app/controllers/api/v0/pages_controller.rb'
 
 Layout/AlignParameters:
   EnforcedStyle: with_fixed_indentation
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -19,6 +19,8 @@ Lint/ParenthesesAsGroupedExpression:
 # Offense count: 16
 Metrics/AbcSize:
   Max: 77
+  Exclude:
+    - 'app/controllers/api/v0/pages_controller.rb'
 
 # Offense count: 5
 # Configuration parameters: CountComments, ExcludedMethods.
@@ -38,11 +40,13 @@ Metrics/LineLength:
 # Offense count: 28
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 78
+  Max: 80
 
 # Offense count: 4
 Metrics/PerceivedComplexity:
   Max: 12
+  Exclude:
+    - 'app/controllers/api/v0/pages_controller.rb'
 
 # Offense count: 7
 # Cop supports --auto-correct.
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -4,11 +4,7 @@ class Api::V0::ApiController < ApplicationController
   before_action :set_environment_header
 
   rescue_from StandardError, with: :render_errors if Rails.env.production?
-  rescue_from Api::NotImplementedError, with: :render_errors
-  rescue_from Api::InputError, with: :render_errors
-  rescue_from Api::DynamicError, with: :render_errors
-  rescue_from Api::AuthorizationError, with: :render_errors
-  rescue_from Api::ResourceExistsError, with: :render_errors
+  rescue_from Api::ApiError, with: :render_errors
 
   rescue_from ActiveRecord::RecordInvalid, with: :render_errors
   rescue_from ActiveModel::ValidationError do |error|
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -3,7 +3,7 @@ class Api::V0::ApiController < ApplicationController
   before_action :require_authentication!
   before_action :set_environment_header
 
-  rescue_from StandardError, with: :render_errors if Rails.env.production? || Rails.env.test?
+  rescue_from StandardError, with: :render_errors if Rails.env.production?
   rescue_from Api::NotImplementedError, with: :render_errors
   rescue_from Api::InputError, with: :render_errors
   rescue_from Api::DynamicError, with: :render_errors
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -8,6 +8,7 @@ class Api::V0::ApiController < ApplicationController
   rescue_from Api::InputError, with: :render_errors
   rescue_from Api::DynamicError, with: :render_errors
   rescue_from Api::AuthorizationError, with: :render_errors
+  rescue_from Api::ResourceExistsError, with: :render_errors
 
   rescue_from ActiveRecord::RecordInvalid, with: :render_errors
   rescue_from ActiveModel::ValidationError do |error|
--- a/app/controllers/api/v0/maintainers_controller.rb
+++ b/app/controllers/api/v0/maintainers_controller.rb
@@ -25,6 +25,48 @@ class Api::V0::MaintainersController < Api::V0::ApiController
     }
   end
 
+  def create
+    data = JSON.parse(request.body.read)
+    if data['uuid'].nil? && data['name'].nil?
+      raise Api::InputError, 'You must specify either a `uuid` or `name` for the maintainer to add.'
+    end
+
+    @maintainer =
+      if page
+        if data['uuid']
+          page.add_maintainer(Maintainer.find(data['uuid']))
+        else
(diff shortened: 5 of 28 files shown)


Mr0grog:
Addendum to (4):

Should we leave maintainers and tags off the page object entirely? (So you only get them from `GET /pages/{uuid}/tags`, for example.) Leave them off only on the list endpoint? Leave them off unless requesting with a `?include_maintainers&include_tags` query argument, like we do versions? Or just keep them around like they currently are in this PR?

danielballan:
> 1. Do we need all the extra routes for listing/renaming/adding maintainers and tags as above?

That depends on the answer to (4). I think we'll probably _eventually_ want on these either way, but I can see the argument for waiting a bit until we are more confident that we have the right model.

> 2. Tag names are not case-sensitive (for querying, that is; the response shows the case they were originally set in), but maintainer names are. This made sense to me when I wrote it, but in hindsight, I’m not so sure.

For tag names, case-insensitivity makes sense. For agencies, I think I would go with the same model that Gmail takes with casing and dots: the model should store and normalize to a _preferred_ capitalization (NASA) but when it comes to defining uniqueness, it should not allow ambiguous variants (NAsA) to be registered as separate entries in the the database.

> 3. Unlike versions, the page listing always includes associated maintainers and tags. I think that’s OK because we probably don’t expect there to be a lot of them, but who knows, that could be a terrible assumption.

It is not obvious to me that this is wrong, and it does not sound too hard to walk back if we discover that it is wrong, and it makes thing convenient for now. Let's keep it.

> 4. Maintainers and tags are each exposed on page records as the full serialization of their object [...]

These objects have a fixed set of fields, and that set is not large. I think including the full object is the right call.

> 5. [Is `tagged_at` a good idea?]

Good idea! Yes, I think `tagged_at` is the useful thing to expose here. Should we do the same for maintainers (`assigned_at`)? I can't think of a situation where the time a tag or maintainer was _created_ would be useful to know. It's the association that matters.

> 5. [Should we make it possible to query tags by AND and OR?]

In a first past, I think whatever Rails makes easy is the right way to go.

> 6. Maintainers and tags both use UUIDs. At this point, I feel like the few places I used numeric IDs make things confusing, so I decided not to do that here, even though they don’t remotely have the constraints around quantity of values or ability to create an ID without already knowing what’s in the DB that pages and versions do.

I see. I think numeric IDs are only preferable when both (1) it is possible to safely avoid collisions without great effort and (2) there is a semantic ordering that is useful to represent. I agree that (1) is true in this case, but (2) is not, so UUIDs are fine.

Mr0grog:
> I think we'll probably eventually want on these [maintainers/tags routes] either way, but I can see the argument for waiting a bit until we are more confident that we have the right model.

I’ll try and implement at least the various `GET` routes—should actually be pretty straightforward. No fancy querying or filtering for the scope of this PR.

> For agencies, I think I would go with the same model that Gmail takes with casing and dots: the model should store and normalize to a preferred capitalization (NASA) but when it comes to defining uniqueness, it should not allow ambiguous variants (NAsA) to be registered as separate entries in the the database.

Cool, this is basically how tags work and how I was starting to feel I should’ve done maintainers. **One further question here, though:** the only variation we are disregarding is upper/lower case (and doing it through the convenient magic of the [`citext`](https://www.postgresql.org/docs/9.6/static/citext.html) Postgres column type). Are there other characters or variations you think we should be disregarding? (In which case I’ll need to implement a little less lazily.)

> Yes, I think `tagged_at` is the useful thing to expose here.

OK, I’ll poke at doing this. It’s a little complicated because what we’ll be returning isn’t an actual object, but an amalgam of the object (tag/maintainer) and the N:M relation table (tagging/pages_maintainers):

```js
{
  "name": "From the object itself",
  "uuid": "From the object itself",
  "tagged_at/assigned_at": "From the relationship table (and a differently named field under the hood—created_at)"
}
```

^ should be totally doable, but this does make it kinda special.

> In a first pass, I think whatever [AND/OR querying] Rails makes easy is the right way to go.

Well, it’s more what SQL makes easy, but 👍. I’ll leave as is. (OR-based query)

(And as a nice bonus, the OR-ing means we can query all the sites someone is assigned in one go, instead of one request per site 😄)

Mr0grog:
FYI, since this is still a recently posted PR, I’m going to actually modify the original migrations for the bits that affect the schema here (so we can keep the chain of migrations *relatively* clean and sane). So if you’ve pulled this branch down and done migrations, you’ll want to roll your DB back before updating.

danielballan:
> Are there other characters or variations you think we should be disregarding? (In which case I’ll need to implement a little less lazily.)

I don't think so, especially as `citext` makes keeps this simple.

> (And as a nice bonus, the OR-ing means we can query all the sites someone is assigned in one go, instead of one request per site 😄)

Awesome. It's good to have a concrete, need-based reason to back this up.

> Hmmmm, actually another question: I don’t think tagged_at/assigned_at should be created_at (too confusing), but is it good that they are different and contextually named, or should they be the same, e.g. assigned_at or associated_at or attached_at or something?

Good point, consistency would be convenient here. I think `assigned_at` works fine for both. I would be equally happy with the other names you proposed.

Mr0grog:
OK, maintainers and tags now have their `created_at`/`updated_at` fields removed and replaced with `assigned_at` in API responses for Pages.

Mr0grog:
AND THAT’S IT! This should be done (sans any new review feedback). This turned out to be much more complex than anticipated (mainly dealing with that horrible `lightweight_query` thing I made and adding like 10 new API endpoints).

Ready for a final review by anyone interested, but probably mainly @danielballan.

lightandluck:
I'm glad you're on our side! 💯

danielballan:
I intend to give this a detailed review tomorrow.

db/migrate/20180212043742_dedupe_pages_with_multiple_agencies_or_sites.rb
```diff
@@ -0,0 +1,45 @@
+class DedupePagesWithMultipleAgenciesOrSites < ActiveRecord::Migration[5.1]
+  # Find all pages sharing the same URL and de-duplicate them by copying the
```
danielballan: :+1:

Mr0grog:
:D

danielballan:
Good to merge once create else path is fixed as discussed on Slack.

Mr0grog:
Updated that create else path:

- When creating a new maintainer (`POST /maintainers`), it’s a `409 Conflict` error to send a `name` that matches an existing Maintainer.

- When adding a maintainer to a page (`POST /pages/{page id}/maintainers`), it’s OK to send details that specify a maintainer that is already on the page. It just returns the one already there (same as before). However, if the details you sent imply a new Maintainer record needs to be created, you get a `409 Conflict` error if it would conflict with an existing Maintainer. In practice, that means you posted:

    ```js
    {
      "name": "some existing maintainer name",
      "parent_uuid": "something different from the existing maintainer with a matching name"
    }
    ```

PR #225, opened 2018-02-18
Add sorting via `?sort=` query parameter

### LINKED ISSUE #165: Add a parameter to control ordering on list requests

Mr0grog:
Fixes #165.

The overall approach here is inspired by @tylerferraro’s original PR from a few months ago. This adds sorting to all list requests with a `?sort=field:order,field:order,...` query parameter. For example:

```
GET /api/v0/pages?sort=title:asc,updated_at:desc
```

Sorting happens in order, so results are sorted by the first field, then, only if that field is equal, by the second field, and so on.

Open questions:

1. Is `sort` the right name here? (vs. `order`? or something else?)
2. Is comma-separated the right thing, or should this be `?sort[]=title:asc&sort[]=updated_at:desc`?
3. Is `field:order` the right way to go here? Seems like that sounded best to everyone earlier on.

### DIFF
--- a/app/controllers/api/v0/annotations_controller.rb
+++ b/app/controllers/api/v0/annotations_controller.rb
@@ -1,9 +1,11 @@
 class Api::V0::AnnotationsController < Api::V0::ApiController
+  include SortingConcern
   before_action :set_annotation, only: [:show]
 
   def index
-    paging = pagination(parent_change.annotations)
-    annotations = parent_change.annotations.limit(paging[:chunk_size]).offset(paging[:offset])
+    annotations = sort_using_params(parent_change.annotations)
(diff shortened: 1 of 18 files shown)


Mr0grog:
Rebased to fix merge conflict from hotfix yesterday.

Mr0grog:
If no concerns, I’m going to go ahead and put this in tonight so that other tools can use it.

PR #231, opened 2018-02-20
HOTFIX: querying by capture_time + source_type/hash

Mr0grog:
Querying by capture_time in combination with other version parameters (`hash`, `source_type`) was broken in the maintainers update. In addition, I'm not sure why I flipped it to a left_outer_joins in the first place, which is clearly the wrong join for this particular query. I think this must have been a mistaken thing that got committed when I was trying to get all the various query params to work happily together.


Mr0grog:
(Circle error was an actual DB error on _Circle_, not our code/tests.)

Issue #61, opened 2017-05-31
Ingest the legacy uuids from the old versionista outputter


danielballan:
It would be nice to have a way to associated legacy Annotations, which I assume will be subjected to a lot of analysis, with Versions in our app. Somehow getting the uuids generated by the old outputter (and now stored only in Google Sheets, I think) sounds slightly painful but possible and useful.

Mr0grog:
Since rows in those spreadsheets are really versions, this is should probably be matching Versionista version IDs. We can extract them from:

- In the spreadsheets: the `Last Two - Side by Side` column is always `https://versionista.com/{site}/{page}/{version_id}:0` (`version_id` is universally unique within Versionista, so all the other fields can be ignored)
- In the DB: `version_record.source_metadata. version_id`

The naive way someone could do this now would be to page through all the results of https://web-monitoring-db.herokuapp.com/api/v0/versions?source_type=versionista

If we wanted to better support this, we could:

- Add some indexing on `source_metadata.versionista_id` and allow querying by that field or
- Make public DB exports available (#45)

danielballan:
Using the versionista ID is good enough to support the _ad hoc_ analysis I want to do right now. Once we transition away from versionista, perhaps we should do a one-time update to the database to ingest all these legacy UUIDs and associated Annotations.

Mr0grog:
:+1:

weatherpattern:
As we move forward with having different differs as well, an annotation imported from sheets should have a field indicating that’s where it came from, (Versionista, Scanner, possibly others in the future.)

danielballan:
I don't think we'll need to worry about that when importing sheets of annotations. Each annotation (row in the sheets) already refers to a Version in our database, either by its Versionista ID or by web-monitoring-db UUID or both, and each Version already knows where it came from.

Mr0grog:
This will be a requirement for migrating away from Google sheets for important changes.

Mr0grog:
This is old enough to no longer be relevant. For changes that were ultimately deemed significant, we updated these IDs in a lot of spreadsheets. In older, non-significant sheets, it will continue to be an issue, but that's OK. Nobody is referencing or using those in any practical capacity (or any capacity at all, so far as I am aware).

PR #233, opened 2018-02-23
Add rake task for importing annotations

Mr0grog:
I was working on some other Ruby + Google Sheets code today, so I took a little extra time for this. We can use it to solve #61 (but that issue is about the work of *running* this script, not about the script itself).

Use the `import_annotations_from_sheet` rake task to import all the annotations an analyst has created in a given Google Sheet. This can be used to solve #61.

Arguments are:
1. Google sheet ID, e.g. `1-Rq-AclS2GI_yxLmkYVY7FvTfN21KoJtxXtOXXXXXX`
2. E-mail of user to attribute the annotation to
3. (optional) Name of spreadsheet tabs to import (comma-separated). If unset, all tabs will be imported.
4. (optional) Row to start at (defaults to 7)
5. (optional) Row to end at. If unset, reads all rows.

When done, it'll output summary information of how many rows were added, skipped, or errored across how many tabs.

Example:

```sh
$ rake import_annotations_from_sheet['1-Rq-AclS2GI_yxLmkYVY7FvTfN21KoJtxXtOXXXXXX','someone@example.com']
> Importing spreadsheet tab '1 EPA About OECA'
> Importing spreadsheet tab '1 HHS Climate & Env'
> 
> RESULTS:
> --------
> Created 253 annotations
> Skipped 0 rows
> Errored 0 rows
> In      2 spreadsheet tabs
```

If you set the `VERBOSE` environment variable (to any value) it'll print information about every row that produced an annotation.


Mr0grog:
I think you’ve already thought about this more than I did! I’m not sure about the colons, though—I just tried using one as a tab title and it seemed to work. Maybe we’re talking about different things?

screen shot 2018-02-26 at 9 29 57 pm

danielballan:
I found that claim on [this Microsoft support page](https://support.office.com/en-us/article/rename-a-worksheet-3f1f7148-ee83-404d-8ef0-9ff99fbad1f9):

> Worksheet names cannot: [...] Contain any of the following characters: / \ ? * : [ ]

I guess Google Sheets went their own way on this one. Let's go with commas and assume we won't hit this bug during the finite period we will be using this feature.

Mr0grog:
Oh, that is an interesting difference. I’ll go ahead and merge this, then.

PR #237, opened 2018-02-26
Canonical, SURTified Page URLs

### LINKED ISSUE #185: Handle pages that exist at multiple URLs
There’s a more complex version of this in the future (e.g. tracking when a page has “moved”), but for now, we know that many of the sites we track server the same resource over both `http` and `https` and from `www`, `www[number]`, and no subdomain at all. This was first brought up in edgi-govdata-archiving/web-monitoring#82, but we’re going to solve the narrow version of this we are encountering now with IA data purely in DB, hence this issue.

The basic idea, as discussed with @danielballan:

1. Add a field to the `Page` model that is the [SURT-formatted URL](http://crawler.archive.org/articles/user_manual/glossary.html#surt). (See also https://github.com/internetarchive/surt for example code to work from)
2. When importing a version, we look to see if we already have a [Page record for it](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/jobs/import_versions_job.rb#L102-L116). That’s done with a lookup based on the URL. Now this will be a multi-step process:

    1. Look for a page with an exact URL match
    2. If no match, look for a page with a SURT URL match
    3. If no match, create a Page
    4. If the page record we have has a “less canonical” URL than the version we are importing, change the Page’s URL to the version’s URL. Most to least canonical:
        1. `https://domain`
        2. `https://www.domain`
        3. `https://www[number].domain`
        4. `http://domain`
        5. `http://www.domain`
        6. `http://www[number].domain`

3. Versions should always have `source_metadata.capture_url` set to the *exact URL they were captured from.*

For now, we expect exceptions where *different* content is served at those 4 URLs to be rare. If we find those situations, we can create a new Page record and move the differing versions to it. Because we look for exact matches *before* SURT matches when importing, imports won’t undo that separation.

Mr0grog:
~**Work in progress; do not merge!**~

Add support for canonicalizing URLs with SURT (Internet Archive’s “Sort-friendly URI Reordering Transform”) so that we can support pages whose content may come from more than one exact URL (as is often the case when we try to import data from Internet Archive).

The idea here is that we add a `url_key` field to Pages and, when importing, match pages by that field *when they don’t match by an exact `url`* (this lets us easily disambiguate two pages with the same `url_key` that should actually be separate by simply making a new Page record with an exact URL match).

- [x] Add module for SURT-formatting a URL
- [x] Add `url_key` field to Page model and populate it
- [x] Add `capture_url` field to Version and populate it
- [x] Add import logic that looks for Pages with matching `url_key` when importing
- [x] Dedupe existing pages by `url_key`

Fixes #185.

### DIFF
--- a/test/lib/surt/surt_test.rb
+++ b/test/lib/surt/surt_test.rb
@@ -17,12 +17,12 @@ class SurtTest < ActiveSupport::TestCase
     assert_canonicalized(
       'http://alexa.com/',
       'http://www.alexa.com/',
-      'It did not remove a "www." subdomain'
+      'It failed to remove a "www." subdomain'
     )
     assert_canonicalized(
       'http://archive.org/',
       'http://www34.archive.org/',
-      'It removed a "www[digit]." subdomain'
+      'It failed to remove a "www[digit]." subdomain'
     )
   end
 
@@ -30,17 +30,17 @@ class SurtTest < ActiveSupport::TestCase
     assert_canonicalized(
       'http://archive.org/index.html',
       'http://archive.org/index.html?',
-      'It removed an empty query'
+      'It failed to remove an empty query'
     )
     assert_canonicalized(
       'http://archive.org/index.html?a=b&b=b',
       'http://archive.org/index.html?b=b&a=b',
-      'It alphabetized the query'
+      'It failed to alphabetized the query'
     )
     assert_canonicalized(
       'http://archive.org/index.html?a=b&b=a&b=b',
       'http://archive.org/index.html?b=a&b=b&a=b',
-      'It alphabetized a query with repeat keys'
+      'It failed to alphabetized a query with repeat keys'
     )
   end
 
@@ -82,17 +82,17 @@ class SurtTest < ActiveSupport::TestCase
     assert_canonicalized(
       'http://168.188.99.26',
       'http://168.188.99.26',
-      'It did leave a simple IPv4 alone'
+      'It failed to leave a simple IPv4 alone'
     )
     assert_canonicalized(
       'http://15.0.0.1',
       'http://017.0.0.1',
-      'It did not decode an octal IP'
+      'It failed to decode an octal IP'
     )
     assert_canonicalized(
       'http://195.127.0.11/blah',
       'http://3279880203/blah',
-      'It did not decode a DWORD host'
+      'It failed to decode a DWORD host'
     )
     # Python SURT does some crazy fixing of IPv4 addresses, but I'm not sure
     # it's actually worth supporting here.
@@ -169,7 +169,7 @@ class SurtTest < ActiveSupport::TestCase
     assert_canonicalized(
       'http://xn--n3h.com',
       '☃.com',
-      '?'
+      'It failed to IDNA-encode the host name'
     )
   end
 
@@ -202,7 +202,7 @@ class SurtTest < ActiveSupport::TestCase
     assert_canonicalized(
       'http://gotaport.com:1234/',
       'http://gotaport.com:1234/',
-      'It did not keep non-standard port numbers'
+      'It failed to keep non-standard port numbers'
     )
   end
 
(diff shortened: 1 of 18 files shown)


Mr0grog:
Obviously still very much a work in progress, but the really complicated part is done here. I’ve generally followed the logic from Internet Archive’s [official Python library](https://github.com/internetarchive/surt), but attempted to clean it up and make nearly every transformation configurable (transforms that merely make the URL *valid* are not configurable).

In researching other canonicalization tools, the two that had some potentially useful additional transforms were:
- [`normalize_url`](https://github.com/rwz/normalize_url) (Ruby)
    - It has some extra query parameters it knows to remove, which we might want to add: https://github.com/rwz/normalize_url/blob/master/lib/normalize_url/normalizer.rb#L8-L18
- [`Purell`](https://github.com/PuerkitoBio/purell) (Go) This is what Sentry uses.
    - I added scheme lower-casing from here
    - It can normalize escapes by upper-casing them (but we effectively normalize by unwinding and re-writing all escapes)
    - It can add a trailing slash (SURT only does nothing or removes)
    - It can remove index file names (e.g. `index.html` at the end of a path). This one is pretty interesting, but potentially very different from IA. Also probably very unsafe.

Things we do **not** support Internet Archive’s SURT:
- Reverse DNS lookup
- “Fixing” IPv4 addresses where one of the octets is bigger than an octet (e.g. `http://10.0.258` → `http://10.0.1.2`)

Mr0grog:
So one thing this leaves out: I’m not updating `Page.url` to the most canonical of actual URLs seen so far. I realized that the comparison of canonical-ness wouldn’t really be able to share much with the actual canonicalization and ultimately kind of felt like have the Page’s `url` field change every so often might not be ideal. Let me know if I should go back and try to implement it, though.

Mr0grog:
Otherwise, this is done and ready for review.

Mr0grog:
Rebased to fix a merge conflict. Also added a Heroku release phase for migrations (https://devcenter.heroku.com/articles/release-phase). I can’t believe I didn’t know about that feature!

test/lib/surt/surt_test.rb
```diff
@@ -0,0 +1,297 @@
...
+      'http://archive.org/index.html',
+      'http://archive.org/index.html'
+    )
+  end
+
+  test 'it removes "www[digit?]" subdomains' do
+    assert_canonicalized(
+      'http://alexa.com/',
+      'http://www.alexa.com/',
+      'It did not remove a "www." subdomain'
```
danielballan: Is the "sign" of this message reversed? This message should state the _desired_ behavior, right? I could be confused by I don't think the "sign" of these messages are consistent within the test.

Mr0grog: The standard convention I’ve seen in the Ruby community is to use the message to say what went wrong (e.g. http://guides.rubyonrails.org/testing.html#your-first-failing-test) and use the test name to say what you expect to have happened. So I’ve generally been trying to follow that, but it looks like this is the only I did that way in this set of tests :\  (See, for example, https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/test/models/annotation_test.rb)

Hmmm, which way do you want to go here?

danielballan: That is the convention I am used to as well. What confused me is that the two messages in this test choose opposite ways: 'It did not remove a "www." subdomain' and 'It removed a "www[digit]." subdomain'. Let's aim for the standard convention as you described it above and apply it in this PR. I'm not too concerned about going back and fixing places where this is reversed because the code, at least, is clear, and we have bigger things to do. Sound reasonable? I hope I'm bike-shedding here; if you think I am feel free to say so. :- )

Mr0grog: Huh, on further review, it looks like I was just all over the place here. Half the messages are sensible and half not. Not sure where my brain was when I did those tests :(

Updated them all to be (what failed). Also changed several use the language “it failed to…” since that seems more clear.

Mr0grog:
OK, fixed the test messages and rebased to resolve a merge conflict. Awaiting any more feedback.

danielballan:
LGTM

PR #242, opened 2018-03-07
Add `citext` postgres extension to README

Mr0grog:
@danielballan is this what you were thinking of for `citext`?


danielballan:
Yes. Thanks.

PR #243, opened 2018-03-08
Add extra query params to strip for SURT

Mr0grog:
This is a follow on to #237 to flag some more querystring params for removal. I had meant to do this in that issue (see https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/237#issuecomment-368617677), but then forgot ◔̯◔

This also adds some TODO comments about potential extra features (that are not very high priority right now).


app/lib/surt/canonicalize.rb
```diff
@@ -34,12 +40,25 @@ module Surt::Canonicalize
     /^(.*\/)(\([0-9a-z]{24}\)\/)([^\?]+\.aspx.*)$/i
   ].freeze
 
+  # TODO: should we refactor this into a more readable format? e.g:
```
danielballan: Meh. It's not to difficult to divine the intent here.

Mr0grog: I do worry about ones that check multiple params in combination, e.g:

```
/^(.*)(?:cfid=[^&]+&cftoken=[^&]+)(?:&(.*))?$/i
```

…that are susceptible to missing their targets if they’re out of order. But if so, we’re no worse than archive.org is. In either case, it’s no big deal right now.

PR #244, opened 2018-03-11
Fix ActiveRecord error messages in import job

Mr0grog:
While working on cleaning up the aftermath of edgi-govdata-archiving/web-monitoring-versionista-scraper#71, I realized we weren't surfacing validation errors that were built into our Page models. This adds that, but in the process, I realized we were also handling ActiveRecord::RecordInvalid errors incorrectly. So this fixes that, too.


Mr0grog:
Gonna go ahead and put this in tonight if nobody has concerns.

PR #245, opened 2018-03-12
Remove auto-migrations as part of Heroku release process

Mr0grog:
Lesson learned: this is great for schema migrations, but for large data migrations, it can be a big problem. This was not the brilliant idea I thought it was.


Mr0grog:
Merging so I can push to Heroku :\

PR #251, opened 2018-03-18
Increase redirect limit to 10 when archiving a URL

Mr0grog:
Internet Archive stores all the redirects in the original request, which means they get replayed to us or anyone else who tries to retrieve an IA page. It turns out that sometimes that's a lot of redirects! Increase our limit to 10 for now (the default in HTTParty is 5).

This fixes some issues I hit while working on https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/174.

### DIFF
--- a/lib/archiver/archiver.rb
+++ b/lib/archiver/archiver.rb
@@ -2,6 +2,8 @@ require 'digest'
 require 'httparty'
 
 module Archiver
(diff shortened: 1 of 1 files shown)


Mr0grog:
Will merge this tomorrow if no feedback.

PR #258, opened 2018-03-24
Change fixtures to sort the same across systems

Mr0grog:
AKA fix test failures on some Linuxes.

It turns out Ubuntu desktop vs. Ubuntu server vs. Mac vs. Windows have different system-level libraries for text collation (sorting). What I did not expect was that *Ruby* does not use them, while *Postgres* does. And, most painfully, `.` vs. `5` sorts differently across some of these systems.

What this all adds up to is odd-seeming test failures around sorting on Ubuntu Desktop because we sort pages with the URL `example.com` and `example2.com` and then validate the sort result in Ruby (which doesn't work because Ruby sorts differently from Postgres there). I'm not sure what the “right” answer is here, so I've just modified the fixtures in such a way that we simply don't hit this scenario.

@gblike does this fix it for you? If so I’ll merge.

### DIFF
--- a/test/controllers/api/v0/pages_controller_test.rb
+++ b/test/controllers/api/v0/pages_controller_test.rb
@@ -69,7 +69,7 @@ class Api::V0::PagesControllerTest < ActionDispatch::IntegrationTest
 
   test 'can filter pages by site' do
     sign_in users(:alice)
-    site = 'http://example.com/'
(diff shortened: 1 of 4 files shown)


gblike:
@Mr0grog Indeed it does. The tests all pass with this.

Mr0grog:
Great! Merging.

Issue #158, opened 2017-10-11
Seed script should prompt if Archiver.allowed_hosts is empty


Mr0grog:
The issues around the `ALLOWED_ARCHIVE_HOSTS` env var (and the `Archiver.allowed_hosts` property is is eventually actualized as) are subtle and have been tripped over by many people. (Most recently in #149).

One thing we should do to help is have the seed script check whether `Archiver.allowed_hosts` is empty (and maybe check that it contains the host used for all the seed records: `https://edgi-versionista-archive.s3.amazonaws.com/`) and, if so, print a warning to the console and prompt the user to type ”y/n“ if they want to continue. (Seeding without this set can take a *long* time because you have to download and process a lot of data—especially if not on a very speedy connection.)

PR #263, opened 2018-03-28
Add prompt for seeding database without archive hosts

gblike:
For issue  #158. All manual testing.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,4 +1,20 @@
 # Seed the DB with an admin user and a few version records
+if Archiver.allowed_hosts.empty?
+  puts <<~WARN
+    No Archiver allowed hosts configured. This will result in a large download.
+    You may set this with the ALLOWED_ARCHIVE_HOSTS environment variable (See .env.example).
+
+    Do you want to continue? [Y]/n
+  WARN
+  response = STDIN.gets.strip.downcase
+  response = response.empty? ? 'y' : response
+  if response == 'y'
+    puts 'Proceeding'
+  else
+    abort('Aborting database seeding')
+  end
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,4 +1,20 @@
 # Seed the DB with an admin user and a few version records
+if Archiver.allowed_hosts.empty?
+  puts <<~WARN
+    No Archiver allowed hosts configured. This will result in a large download.
+    You may set this with the ALLOWED_ARCHIVE_HOSTS environment variable (See .env.example).
```
Mr0grog: This is awesome, thank you!

That said, I wonder if we could explain more here, e.g:

> You don't have any allowed hosts configured for Archiver! That means every version of every web page in the seed file will be downloaded to your computer, which could take a while. To avoid this, stop this script and set the ALLOWED_ARCHIVE_HOSTS environment variable (see .env.example).
> 
> (Allowed hosts are hosts that are allowed to remotely store the raw content of a version of a web page on behalf of your database. If a version is stored at an allowed host instead of locally, you'll need to be online in order to view the webpage itself or diffs between version. Metadata is still stored locally in your database for offline use.)

Do you think that’s helpful or just way too verbose? (What you’ve already got is good, too.)

gblike: I think we can put that in the `.env.example` file and really emphasize that someone should go look at it for more information.

Mr0grog:
Sorry I’ve been slow to circle back around and merge this, @gblike! Do you want me to add you to the [contributors list](https://github.com/edgi-govdata-archiving/web-monitoring-db#contributors)?

PR #277, opened 2018-04-13
Add `?create_pages=false` import query argument

### LINKED ISSUE #265: Imports should have a `?create_pages=true|false` querystring argument
## Background

In the new IA bulk import script, we try to be efficient by querying for all snapshots in a given domain for the time period we’re concerned about. Unfortunately, that means we will almost certainly pick up snapshots for pages we don’t care about. We might want those in the future, but not right now. This presents a problem:

- We can’t filter the responses based on exact URL matches because we know multiple URLs correspond to a single page, e.g. `http://www.epa.gov == https://www.epa.gov`

- We are currently filtering on [`url_key`](https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/174/files#diff-b12cfc8d93e2d8ce1cf3f93ea9809cb8R80), but that only works because our canonicalization method happens to be modeled after Internet Archive’s. The problem is that we might have good reason to change our canonicalization in the future or Internet Archive might change theirs. We can’t always expect these to match up — even though it *mostly* works right now.

- We can’t canonicalize the URL we got from IA according to our rules on the fly because the import script is not part of the DB API. Even if they were in the same language, so we could share the code (they aren’t, so we can’t), we wouldn’t have any good guarantee that the version of the import script you’re using matches up with the version of the DB API you are sending data to.

So… I kinda feel like we have to filter out versions and determine whether to add them as part of the actual importing process *here,* inside the DB API.

## Proposal

I propose we add a `?create_pages=true|false` querystring argument that controls whether new page records can be created if the version to be imported doesn’t match up to any existing page. The default behavior would be `true`, which is how the importer currently behaves (if the the version doesn’t correspond to an existing page, we create a new page for it). If `false`, we simply wouldn’t import the version.

It’s a boolean argument, so if there’s no value at all (i.e. no equals sign), that’s equivalent to `true`:

- `?`, `?create_pages`, `?create_pages=true`, `?create_pages=TrUe`, `?create_pages=1` are all `true`
- `?create_pages=false`, `?create_pages=FaLsE`, `?create_pages=0` are all `false`
- Other values are errors

Mr0grog:
To prevent imports from creating completely new pages (when there's a previously unknown URL in the import data), set the `create_pages` query argument to false in `POST /api/v0/imports?create_pages=false`.

Fixes #265.

### DIFF
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -36,7 +36,7 @@ class ImportVersionsJob < ApplicationJob
         messages = error.record.errors.full_messages.join(', ')
         @import.processing_errors << "Row #{row}: #{messages}"
       rescue StandardError => error
-        @import.processing_errors << if Rails.env.development?
+        @import.processing_errors << if Rails.env.development? || Rails.env.test?
                                        "Row #{row}: #{error.message}"
                                      else
                                        "Row #{row}: Unknown error occurred"
@@ -47,7 +47,9 @@ class ImportVersionsJob < ApplicationJob
   end
 
   def import_record(record)
-    page = find_or_create_page_for_record(record)
+    page = page_for_record(record, create: @import.create_pages)
+    return unless page
+
     existing = page.versions.find_by(
       capture_time: record['capture_time'],
       source_type: record['source_type']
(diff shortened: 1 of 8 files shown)


app/jobs/import_versions_job.rb
```diff
@@ -47,7 +47,9 @@ def import_raw_data(raw_data)
   end
 
   def import_record(record)
-    page = find_or_create_page_for_record(record)
+    page = page_for_record(record, create: @import.create_pages)
+    return unless page
```
danielballan: Does it seem potentially useful to issue a log message when a `url` is skipped here? That might be too verbose, depending on how often this happens in practice. Just a thought.

danielballan: How about a separate list, processing_warnings?

The endpoints also seem like a very good idea but not necessarily a blocker for the PR.

Mr0grog: Does 70641ec8a75ce849219ed961d255e5566cf3ace8 seem like a reasonable approach?

Mr0grog:
Approval from @danielballan over other channels; merging.

PR #287, opened 2018-04-18
Add `?skip_unchanged_versions=true` to imports (#266 Primary)

### LINKED ISSUE #266: Add a `?skip_unchanged_versions=true|false` querysting argument to /imports
## Background

When importing data from Internet Archive (or any source, really), we only know about data in the timeframe we are examining. If a data source (like Internet Archive) doesn’t differentiate between versions that are the same and versions that are different from the ones that precede them, we could wind up importing a version to our database that is no different from the latest version we already have. This is undesirable when we are *generally* trying to only store versions that differ from the version before them (in the long run, that might be worth revisiting, but for now, it’s what we’re doing). We don’t make any guarantees about that, but we try to make sure it’s true in most cases.

For example, suppose our database has versions of the page `http://ugabooga.com/whiz-what-wow.html` like:

| capture_time | hash |
| ------------ | ---- |
| 2018-01-03   | abc  |
| 2018-01-05   | def  |
| 2018-01-20   | ghi  |

And Internet Archive captures a snapshot of that page every day. If the page never changes after 2018-01-20, but we run a script on 2018-01-21 to import new versions, we’ll bet back a version record from Internet Archive like:

| capture_time | hash |
| ------------ | ---- |
| 2018-01-21   | ghi  |

It’s exactly the same content as the latest version we already have (it has the same hash), but looks like a new record because it was captured on a newer date.

At current, we will store this version, which means our API will show two consecutive versions that are exactly the same :(

## Proposal

I propose we add a querystring argument to `POST /imports` called `?skip_unchanged_versions` with a boolean value. If set to `true`, the importer would not import versions of a page that have the same hash as the version captured immediately before them (*from the same source*).

The default value for it would be `false` — matching existing behavior, but also playing it safe. If `true`, we have to guarantee that the versions in the import request *are in capture_time order* (otherwise we might skip them incorrectly).

## Alternatives

We might start storing all snapshots of a page (regardless of whether they are different from the previous), but add a special field named `changed` (or `is_changed` or `is_different` or something similar). That field would be a boolean indicating whether that version’s hash matches the previous version’s hash, and we’d use it when querying to determine whether to show a version.

This is a lot more complicated than simply not importing things:

- We have to make sure the special database field stays up-to-date and accurate whenever adjacent versions are added
- We have to make sure all our queries correctly check that field so they are showing the right things
- It potentially has much higher storage costs (because we’ll store a *lot* more versions)

On the other hand, it’s also a lot more flexible (and safe). ¯\\\_(ツ)\_/¯

/cc @danielballan @weatherpattern @lightandluck 

*File under: this project is slowly becoming a miniaturized archive.org*

Mr0grog:
Fixes #266 via the primary proposal

NOTE: this approach means we will *occasionally* miss versions or capture versions we should not have when they are received or processed out of order.

Mainly for comparison with #279.

### DIFF
--- a/app/controllers/api/v0/imports_controller.rb
+++ b/app/controllers/api/v0/imports_controller.rb
@@ -16,7 +16,8 @@ class Api::V0::ImportsController < Api::V0::ApiController
     @import = Import.create_with_data({
       user: current_user,
       update_behavior: update_behavior,
-      create_pages: boolean_param(:create_pages, default: true)
+      create_pages: boolean_param(:create_pages, default: true),
+      skip_unchanged_versions: boolean_param(:skip_unchanged_versions)
     }, request.body)
     ImportVersionsJob.perform_later(@import)
     show
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -76,6 +76,11 @@ class ImportVersionsJob < ApplicationJob
       end
     end
 
+    if @import.skip_unchanged_versions? && version_changed?(version)
+      warn "Skipped version identical to previous. URL: #{page.url}, capture_time: #{version.capture_time}, source_type: #{version.source_type}"
+      return
+    end
+
     version.validate!
     version.save
   end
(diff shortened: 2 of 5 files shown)


danielballan:
Note: I reviewed this in the context of #288, which is the winner in my opinion.

Mr0grog:
This was incorporated into #288 before it was merged, so it’s already effectively been merged. Closing.

PR #288, opened 2018-04-18
Or, you know, just do all the approaches for #266

### LINKED ISSUE #266: Add a `?skip_unchanged_versions=true|false` querysting argument to /imports

Mr0grog:
I guess it’s not like they really conflict.

Fixes #266 ALL THE WAYS

### DIFF
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -77,6 +77,7 @@ class ImportVersionsJob < ApplicationJob
     end
 
     version.validate!
+    version.update_different_attribute
     version.save
   end
 
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -77,7 +77,7 @@ class ImportVersionsJob < ApplicationJob
     end
 
     version.validate!
-    version.update_different_attribute
+    version.update_different_attribute(save: false)
     version.save
   end
 
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -76,6 +76,11 @@ class ImportVersionsJob < ApplicationJob
       end
     end
 
+    if @import.skip_unchanged_versions? && version_changed?(version)
+      warn "Skipped version identical to previous. URL: #{page.url}, capture_time: #{version.capture_time}, source_type: #{version.source_type}"
+      return
+    end
+
     version.validate!
     version.update_different_attribute(save: false)
     version.save
@@ -174,5 +179,17 @@ class ImportVersionsJob < ApplicationJob
       end.join(' or ')
       raise Api::InputError, "`#{field_name}` must be #{names}, not `#{value.class.name}`"
     end
+  end
+
+  def version_changed?(version)
+    return true if version.version_hash.nil?
+
+    previous = version.page.versions
+      .where(source_type: version.source_type)
+      .where('capture_time < ?', version.capture_time)
+      .order(capture_time: :desc)
+      .first
+
+    version.version_hash == previous.try(:version_hash)
   end
 end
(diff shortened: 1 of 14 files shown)


lightandluck:
Looks good as far as I can tell. Read through code and nothing stuck out. Looked more closely at tests and they make sense and passed. 👍

One minor typo, in the commit message for [355b4212](https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/288/commits/355b421231ba5f18fb01e75e7ce09dd44c7b3ad4), we say the new method is `version.update_changed_attribute`, but it ended up being called `version.update_different_attribute`. Not sure if that's actually worth fixing. ¯\_(ツ)_/¯

app/jobs/import_versions_job.rb
```diff
@@ -174,4 +180,16 @@ def validate_kind!(kinds, record, field_name)
...
     end
   end
+
+  def version_changed?(version)
+    return true if version.version_hash.nil?
+
+    previous = version.page.versions
+      .where(source_type: version.source_type)
+      .where('capture_time < ?', version.capture_time)
+      .order(capture_time: :desc)
```
danielballan: I noticed that, in `update_different_attribute`, you use `limit(1).first` which I assume is a performance win. Does that not buy you anything in this case?

lightandluck:
I went ahead and removed `.limit(1)` and updated the docs, if someone wants to approve and merge =D

danielballan:
Thanks!

PR #290, opened 2018-04-18
Offer local dev option with minimal local installation

patcon:
Hey all! I was trying to work at an dev setup process that will onboard 99% of people in maybe 10 minutes. I'm hoping to start pitching at hacknights, so I'm gearing up for that and hoping to have feedback sooner rather than later.

Having onboarded people in-person to a good few projects at hacknights, getting local underlying services (like redis and postgres) running is usually the biggest stumbling block for new people to get to the point of getting things running on their local machine.

Also, as someone with no enough spare space to run docker images, I personally feel the pain of not being able to use that approach.

Could point out this setup path that off-loads most of the setup onto a remote service like heroku?

Also, I notice that we don't use the `bin/setup` script that rails provides. Would we be in favour of using that more in the setup docs?

(I hope to make some pitches at the hacknight and get people running through this process, and getting some feedback.)


Mr0grog:
> Having onboarded people in-person to a good few projects at hacknights, getting local underlying services (like redis and postgres) running is usually the biggest stumbling block for new people to get to the point of getting things running on their local machine.

Yes, making sure you have all the right other services running and that you’ve got them linked up right is probably the hardest part here. It can be tough if you are doing it for the first time. We don’t give especially good guidance.

> Could point out this setup path that off-loads most of the setup onto a remote service like heroku?

I *really* do not like this approach. I get that it makes things simpler to set up (if you are comfortable with Heroku), but I feel very uncomfortable with suggesting a default environment where critical components of the application are non-local, not to mention co-opting Heroku’s service for a free (if really poor performing) cloud-based DB.

If other people disagree, ok. But I’m not personally excited about this.

> Also, I notice that we don't use the bin/setup script that rails provides. Would we be in favour of using that more in the setup docs?

I suppose we could actually utilize this to simplify some things. I don’t think it’s very worthwhile if you are also going add `make`, though (I think the idea in Rails is that you would use `rake` instead of `make`, which is generally what we do, and then you’d need `setup` to get the gems required for your `rake` tasks to load).

bin/setup
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env ruby
+
+require 'dotenv/load'
```
Mr0grog: This isn’t going to work very well (for first time runs) until *after* line 31 where we install gems.

danielballan:
> Having onboarded people in-person to a good few projects at hacknights, getting local underlying services (like redis and postgres) running is usually the biggest stumbling block for new people to get to the point of getting things running on their local machine.

I agree that this is an important problem. It keeps me up at night that getting up and running is so involved. As part of my work on ansible, I am working on a solution that will ask new developers to do only:

```bash
git clone <OUR_TEMPORARILY_PRIVATE_ANSIBLE_REPO>
brew install ansible  # or sudo apt-get install ansible
sudo ansible-playbook -i inventory/local site.yml
# Aim browser to http://localhost:3001
```

and I think that will make a big difference. A crucial feature of an approach like this one is that it uses the same standard tools (homebrew, postgres, etc.) that a more experienced developer would choose. It avoids introducing concepts that are beginner-friendly on the surface but, in the longer term, add additional cognitive overhead and maintenance burden.

I agree with @Mr0grog's points against using Heroku. We have an app that is complex. Hiding that complexity by using a special deployment strategy for newcomers gets them up and running quicker, but in the medium-term makes things _more_ confusing because eventually we have to say "OK, now forget about Heroku -- unlearn what you've learned so far -- and instead learn the actual way we do this...." Our goal should be to be getting someone all the way to becoming a regular contributor, which is not necessarily the same as getting someone to running the app along the fastest path possible.

weatherpattern:
>  Our goal should be to be getting someone all the way to becoming a regular contributor, which is not necessarily the same as getting someone to running the app along the fastest path possible.

I think @danielballan makes a very good point here, for this to work we need to get a developer onboard (paid or volunteer) who has the capacity to manager that transition from heroku to a local install. 
I recommend getting the Ansible deployment set up and improved documentation out (both in active progress) and reassess.

bin/setup
patcon: If we want to salvage using `bin/setup`, could we not rescue below the `bundle check` like:

```rb
begin 
  require 'dotenv/load'
rescue LoadError
end
```

danielballan:
[Derailing into a philosophical discussion here, just for the fun of it....]

> By the time they're at the deeper level, they're no longer the newcomer on their first install, and so the landscape of what you can ask/expect of this person is totally different. First impressions count, and an "ASAP Local" philosophy still personally wins me over :)

You make a well-considered point. I give this some thought too (mostly [in my day-job work](https://nsls-ii.github.io/tutorials.html)) and I usually speak for the side of the argument that you are taking here. I don't disagree with anything you've said, but I think that it's also important to keep some consistency in tooling and presentation. Broad speaking, if the tooling for the installation and the basic usage examples are _very_ specialized for that first impression, users and contributors might get a false sense of what they are getting into, causing them to misjudge whether a project is a good fit for their needs/interests or, as I suggested in a previous comment, to become confused as they learn more. Some specialization for that first impression is warranted, but I wouldn't agree with a maximalist position.

patcon:
Thanks all! Sorry for the noise

PR #293, opened 2018-04-19
Add logging to seed data import

patcon:
Importing so much data slowly was a bit confusing on the first set-up. I realize I've done this incorrectly, but I'd love feedback on logging the proper way from the seeds task :)

### DIFF
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -27,7 +27,8 @@ class ImportVersionsJob < ApplicationJob
     total_rows = raw_data.split("\n").length
     each_json_line(raw_data) do |record, row|
       begin
-        Rails.logger.info("Importing row #{row+1}/#{total_rows}...") if ((row+1) % 25 == 0)
+        i = row + 1
+        Rails.logger.info("Importing row #{i}/#{total_rows}...") if (i % 25).zero?
         import_record(record)
       rescue Api::ApiError => error
         @import.processing_errors << "Row #{row}: #{error.message}"
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -27,7 +27,7 @@ class ImportVersionsJob < ApplicationJob
     total_rows = raw_data.split("\n").length
     each_json_line(raw_data) do |record, row|
       begin
-        puts "Importing row #{row+1}/#{total_rows}..." if ((row+1) % 25 == 0)
+        Rails.logger.info("Importing row #{row+1}/#{total_rows}...") if ((row+1) % 25 == 0)
         import_record(record)
       rescue Api::ApiError => error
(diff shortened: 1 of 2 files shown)


patcon:
And the thought around exposing progress is that, if people know how long it's taking on their connection, then they can choose to find the file and delete some rows to make it manageable :tada:

app/jobs/import_versions_job.rb
```diff
@@ -24,8 +24,10 @@ def perform(import)
   end
 
   def import_raw_data(raw_data)
+    total_rows = raw_data.split("\n").length
     each_json_line(raw_data) do |record, row|
       begin
+        puts "Importing row #{row+1}/#{total_rows}..." if ((row+1) % 25 == 0)
```
Mr0grog: `puts` is *not ok* in this context. Please use `Rails.logger.[whatever log level is appropriate]`. This needs to be able to function properly for normal imports when the application is running.

FYI, the reason it’s written without a total was so that it would be easy to transition to a streaming, lower-memory model (which necessarily has the caveat that you don’t know how long the stream is until you are done). Most clients in practice are nice to the app and avoid imports so big that we’ve needed to solve that, though, so I guess this is ok.

Mr0grog: Also, the whole import can be a single JSON file instead of a newline-delimited JSON stream, so `raw_data.split('\n').length` is not an accurate count of records.

patcon: Sorry, I was trying to admit with the "poor [man's] logging" commit msg that this was ugly and wrong, but just trying to show an aspiring user experience that I thought might be more friendly to show progress on 15+ minutes of silent output :)

I messed around with `Rails.logger.info` for 15 minutes, but it seemed to be shuttling output into `log/development.log` or some such place that wasn't obvious, so I thought someone who understood rails better could give the quick advice on how to do this right. Halp? (Any advice on custom loggers -- if that's the right approach -- would be sincerely appreciated!)

patcon:
Ok, should be fixed :)

Mr0grog:
Ah, I see what you meant about backscroll here, @patcon. This is also going to add a lot of noise to our normal operational logs, so maybe using the logger was bad advice on my part :\

What about if we did something like [this, with an additional argument and a TTY check](https://github.com/edgi-govdata-archiving/web-monitoring-db/compare/fancier-seed-import-logging), or even fancier, like [this, where you can pass in a proc to handle progress reporting](https://github.com/edgi-govdata-archiving/web-monitoring-db/compare/extra-fancier-seed-import-logging)? Both of those also result in a single, self-updating log line so there’s less painful backscroll.

Mr0grog:
@patcon did you have a chance to take a look at either of those in the past week?

Mr0grog:
I think this is still important to finish and address. I guess it’s on me to do so.

Mr0grog:
This was superseded by #529.

PR #294, opened 2018-04-19
Read database credentials from dotenv

patcon:
I'm one of those people on an old platform with postgres defaulting to be on non-standard ports :)

I have to adjust `config/database.yml` and so my workspace is always dirty.

Can we move database config into the .env  as DATABASE_URL file so that people can do whatever they please locally, and it will stay out of git-space? This has the nice perk of being more aligned with how we do it on prod.

Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
           PGHOST: 127.0.0.1
           PGUSER: root
           HOST_URL: 'web-monitoring-db.test'
-          DATABASE_URL_TEST: 'postgres://root@localhost:5432/circle-test_test'
+          DATABASE_TEST_URL: 'postgres://root@localhost:5432/circle-test_test'
           ALLOWED_ARCHIVE_HOSTS: 'https://edgi-versionista-archive.s3.amazonaws.com/ https://test-bucket.s3.amazonaws.com/'
       - image: circleci/postgres:9.5-alpine
         environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ jobs:
           PGHOST: 127.0.0.1
           PGUSER: root
           HOST_URL: 'web-monitoring-db.test'
+          DATABASE_URL_TEST: 'postgres://root@localhost:5432/circle-test_test'
           ALLOWED_ARCHIVE_HOSTS: 'https://edgi-versionista-archive.s3.amazonaws.com/ https://test-bucket.s3.amazonaws.com/'
       - image: circleci/postgres:9.5-alpine
         environment:
--- a/.env.example
+++ b/.env.example
@@ -4,7 +4,8 @@ HOST_URL='web-monitoring-db.dev'
 
 # OPTIONAL: only set this if your database is at a different location than
 # specified in config/database.yml
-# DATABASE_URL=postgres://user:password@localhost:5432/db-name
+DATABASE_URL_DEV=postgres://localhost:5432/web-monitoring-db_development
+DATABASE_URL_TEST=postgres://localhost:5432/web-monitoring-db_test
 
 # OPTIONAL: only set this if your Redis is at a non-standard location
 # REDIS_URL=redis://user:password@localhost:6379
--- a/.env.example
+++ b/.env.example
@@ -4,8 +4,8 @@ HOST_URL='web-monitoring-db.dev'
 
 # OPTIONAL: only set this if your database is at a different location than
 # specified in config/database.yml
-DATABASE_URL_DEV=postgres://localhost:5432/web-monitoring-db_development
-DATABASE_URL_TEST=postgres://localhost:5432/web-monitoring-db_test
+DATABASE_DEV_URL=postgres://localhost:5432/web-monitoring-db_development
+DATABASE_TEST_URL=postgres://localhost:5432/web-monitoring-db_test
 
 # OPTIONAL: only set this if your Redis is at a non-standard location
 # REDIS_URL=redis://user:password@localhost:6379
--- a/config/database.yml
+++ b/config/database.yml
@@ -24,7 +24,7 @@ default: &default
 
 development:
   <<: *default
-  url: <%= ENV['DATABASE_URL_DEV'] %>
+  url: <%= ENV['DATABASE_DEV_URL'] %>
 
   # The specified database role being used to connect to postgres.
   # To create additional roles in postgres see `$ createuser --help`.
@@ -58,7 +58,7 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  url: <%= ENV['DATABASE_URL_TEST'] %>
+  url: <%= ENV['DATABASE_TEST_URL'] %>
 
 # As with config/secrets.yml, you never want to store sensitive information,
 # like your database password, in your source code. If your source code is
--- a/config/database.yml
+++ b/config/database.yml
@@ -24,7 +24,7 @@ default: &default
 
 development:
   <<: *default
-  database: web-monitoring-db_development
+  url: <%= ENV['DATABASE_URL_DEV'] %>
 
   # The specified database role being used to connect to postgres.
   # To create additional roles in postgres see `$ createuser --help`.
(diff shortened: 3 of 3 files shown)


Mr0grog:
Is there a reason adding a `.env.test` file doesn’t solve this problem just as well? (Note you might have to run `spring stop` after adding/updating it, same as the `.env` file.)

Mr0grog:
(Obviously the README doesn’t mention `.env.[environment name]` files, so maybe that’s what really needs fixing here.)

Mr0grog:
Docs, for anyone who was unaware: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use

patcon:
Sorry, I'm confused. Can you explain how you're imagining `.env.[environment]` would apply here? Like one example of the equivalent `.env.test` would help me understand what you're seeing. (Perhaps we're seeing a different "point" to what this PR is about :) )

Mr0grog:
So if you have a funky database URL, like maybe a custom port, you might have…

`.env` file:

```sh
DATABASE_URL='postgres://localhost:2345/web-monitoring-db_development'
# All your other config vars, e.g:
ALLOWED_ARCHIVE_HOSTS='https://edgi-versionista-archive.s3.amazonaws.com/'
HOST_URL='localhost:3000'
```

`.env.test` file:

```sh
# Just the config vars that are custom to your test environment
# This gets layered atop what’s in `.env`
DATABASE_URL='postgres://localhost:2345/web-monitoring-db_test'
```

…and no changes to the source code or other files are necessary. This also means you can change all kinds of other things in your test environment — not just database config — like `MAIL_SENDER` or `MAX_COLLECTION_PAGE_SIZE` or `AWS_ACCESS_KEY_ID`.

Mr0grog:
@patcon does this work for you? Am I misunderstanding what you were trying to get at here?

Mr0grog:
@patcon just checking in again. It’s been almost two weeks since I posted that example; does it address what you were looking for? Should we switch gears here and document that rather than changing code?

jsnshrmn:
I'm running into this issue in my dev setup too. @Mr0grog I'll let you know if your suggestion cures the issue.

jsnshrmn:
So, I followed through the current readme for configuring things via .env, and also took your suggestion for creating environment-specific overlays. So I have .env.development and .env.test.

Having done that, I'm still not able to actually configure my dev and test database connections as desired since those environments are not configured to pick up all of the same environment variables as production.

Since these environment variables are the documented interface for configuration in production, then I expect to have them in dev and test as well. This also means that we could actually leave authentication on in dev, which would narrow the delta between the environments.

Rather than namespacing out separate variables for the three environments, I think we can just leverage the fact that rails knows its execution environment, I think we should do as @Mr0grog suggests and just use environment-specific values.

But for that to really work railsconfig needs to actually pull in the same variables in all three environments.

Mr0grog:
> Having done that, I'm still not able to actually configure my dev and test database connections as desired since those environments are not configured to pick up all of the same environment variables as production.

What environment variables are they not picking up? I don’t think there are any variables that are only picked up in certain environments (well, with the exception of DATABASE_RDS, which only exists to aid our transition from Heroku Postgres to RDS).

Mr0grog:
> So I have .env.development and .env.test

As a side-note, do you have anything where your dev environment conflicts with test and can’t be overridden? I would think it’d be simpler to do `.env` and `.env.test`, so the latter only has to override the variable(s) it cares about (probably just `DATABASE_URL`).

jsnshrmn:
I'm having particular trouble with the database password, which is part of DATABASE_URL. DATABASE_URL does seem to be getting passed in, but the password within isn't actually getting used in the connection negotiation. I tried also setting WEB-MONITORING-DB_DATABASE_PASSWORD, but that doesn't seem to be doing anything either. I'd expect to be able to pass in an arbirtrary postres role, password, and db name and have it be happy.

jsnshrmn:
> So I have .env.development and .env.test

Yeah, this was me continuing to search for a larger hammer. I have .env as well, which is currently identical to .env.development.

FWIW setting pg_hba to just trust everything makes this issue go away, but I'm trying to configure this thing more properly, as that's only acceptable in a local environment.

Mr0grog:
> `DATABASE_URL` does seem to be getting passed in, but the password within isn't actually getting used in the connection negotiation

Are you seeing this specifically when running `rake` commands, or in general? If you environment is development, all the `rake db:*` commands do a dirty thing and run *twice,* once in your current env (development), and again in test, but they'll be running with ENV and configuration that was build for the dev environment (because that's the environment the process launched in). So commands will run successfully for dev, but when run a second time for test, they fail. If that’s your situation, you can safely ignore the exception and just run the command again with `RAILS_ENV=test rake db:whatever`.

If you’re aren’t having a `rake db:*` specific problem, though, something else bad is happening here.

jsnshrmn:
Hmm, yeah I've got something hinky going on. I woudn't expect rails to touch the test environment when I do a db:create or db:schema:load, but I'm still having problems. I've been trying to run these via ansible, and I think that's where I'm getting bogged down. I'll come back to this tomorrow and try this by hand again, as that's where we're hoping for the phase 0 automation to stop anyway.

jsnshrmn:
Good to know. I have recent ruby experience, but I haven't touched rails in a decade. Clearly I'll have some catching up to do if I want to get very far into this application.

Having said that, I've verified that for whatever reason, DATABASE_URL is definitely getting ignored in my development environment.

I did some experimentation in a live shell.

- I changed the username in the connection string to a nonexistent pg role, and it still executed under my rails user as the pg role.
- When I set PGPASSWORD in my .env files, password authentication works (with the rails user as the pg role).
- I added ```url: <%= ENV['DATABASE_URL'] %>``` under the development railsconfig but it didn't seem to make any difference.

jsnshrmn:
In this environment I haven't gone past db:setup and fired up a server console yet.
```
$ rbenv exec bundle exec spring stop
Spring is not running
```

jsnshrmn:
rerunning setup with DATABASE_URL set in .env

```$ rbenv exec bundle exec rake db:setup
Database 'web-monitoring-db_development' already exists
fe_sendauth: no password supplied
Couldn't create database for {"adapter"=>"postgresql", "encoding"=>"unicode", "pool"=>5, "host"=>"localhost", "database"=>"web-monitoring-db_test"}
rake aborted!
PG::ConnectionBad: fe_sendauth: no password supplied
/home/rails/.rbenv/versions/2.4.1/bin/bundle:22:in `load'
/home/rails/.rbenv/versions/2.4.1/bin/bundle:22:in `<main>'
Tasks: TOP => db:setup => db:schema:load_if_ruby => db:create
(See full trace by running task with --trace)```

rerunning setup with PGPASSWORD set in .env
```
$ rbenv exec bundle exec rake db:setup
Database 'web-monitoring-db_development' already exists
Database 'web-monitoring-db_test' already exists
-- enable_extension("plpgsql")
   -> 0.0106s
-- enable_extension("citext")
   -> 0.0022s
-- enable_extension("pgcrypto")
   -> 0.0023s
-- enable_extension("uuid-ossp")
   -> 0.0027s
-- create_table("annotations", {:primary_key=>"uuid", :id=>:uuid, :default=>#<Proc:0x0055847671b128@/srv/checkout/web-monitoring-db/db/schema.rb:21 (lambda)>, :force=>:cascade})
   -> 0.0145s
...
------------------------------------------------------------
Admin user created with e-mail: seed-admin@example.com and password: PASSWORD
------------------------------------------------------------
...
```

jsnshrmn:
aha! I forgot to swap the railsconfig for the test environment over to use DATABASE_URL too, due to the aformentioned running rake db:* firing in test from the dev process. So I think DATABASE_URL | current hardcoded behavior in database.yml will make this happy for my purposes and preserve current behavior. PR forthcoming

jsnshrmn:
I went ahead and opened and self-assigned an issue for this, since this PR doesn't have an issue attached to  it.
https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/358

Mr0grog:
OK, we’ve gone with #363 instead. Thanks so much for starting the conversation on this, @patcon.

PR #295, opened 2018-04-19
Compress some README steps into bin/setup

patcon:
Dependant on https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/294


README.md
```diff
@@ -46,45 +46,27 @@ It’s a Rails app that:
...
+    <sub>Alternatively, for manual postgres setup, [see below](#manual-postgres-setup).</sub>
 
     If you're getting error such as `FATAL: role "user" doesn't exist. Couldn't create database.` check [troubleshooting](#troubleshooting) below.
 
-10. Start the server!
+7. Start the server!
 
     ```sh
-    $ bundle exec rails server
+    $ bin/rails server
```
Mr0grog: Technically, this should still `bundle exec rails server` or `bundle exec bin/rails server`. A lot of the time, skipping the `bundle exec` part will work fine, but if the app is not compatible with a newer version of a gem you might have (e.g. because you are running multiple apps using the same version of Ruby or maybe different branches use different gems), you can have problems. `bundle exec` is what makes sure the right versions of each gem get loaded.

README.md
```diff
@@ -46,45 +46,27 @@ It’s a Rails app that:
...
     ```
 
-7. Wherever you cloned the repo, go to that directory and install dependencies:
+    This will set up and populate a database with sample data and an admin user. (Read the output for details!)
 
-    ```sh
-    $ bundle install --without production
-    ```
-
-8. Copy the .env.example file to .env - this allows for easy configuration locally.
```
Mr0grog: I feel like we still need a line that at least explains this is what you modify to change your local configuration. That’s an important thing that people need to know how to do.

db/seeds.rb
```diff
@@ -16,8 +16,8 @@
 end
 
 admin = User.find_or_create_by(admin: true) do |user|
-  password = 'PASSWORD'
-  user.email = 'seed-admin@example.com'
+  password = 'someones_password'
+  user.email = 'someone@example.com'
```
Mr0grog: Why did you change the default e-mail and password?

db/seeds.rb
```diff
@@ -35,3 +35,7 @@
 FileStorage.default = FileStorage::LocalFile.new(path: Rails.root.join('db'))
 ImportVersionsJob.perform_now(import)
 FileStorage.default = fs_default
+
+puts "\n\n------------------------------------------------------------"
+puts "Login in with these admin credentials: #{admin.email} / #{admin.password}"
+puts "------------------------------------------------------------\n\n"
```
Mr0grog: Hmmm. A few concerns here:

1. `admin.password` will always return `nil` (it’s just a setter that eventually modifies `encrypted_password`), so this line won’t give you useful credentials.
2. At this point, there’s actually no way to get the password, so you actually *can’t* print usable credentials here (we only store the hashed password).
3. If a user was created, we have sort-of-duplicated output (from line 24), which is weird/confusing.
4. Do we really need to print anything here if they already had a user present and we didn’t actually create one?

db/seeds.rb
patcon: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/d06e93f80c0326ccb2b47aa7c8d0ff980e293905/app/views/home/index.html.erb#L14-L19

As in, let's just keep these in sync, so it's less noise :)

db/seeds.rb
patcon: Ah sorry, I could've sworn I saw this work. Definitely not committed to this way I've done it.

As for why it's helpful here: We have another PR that prints feedback on the sometimes long/silent process of importing: https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/293

I felt that any credentials should be printed at the end of that whole process, when control is kicked back to user. But yes, only one place is necessary.

README.md
patcon: `config/boot.rb` loaded within the default `bin/setup` takes care of this i believe:
https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/config/boot.rb

As far as I can tell, `bin/setup` is meant to be all that's needed

Example of another big project that just uses `bin/setup`:
https://github.com/18F/identity-idp/blob/master/bin/setup

(note that within `bin/rake` is where the boot part is loaded)

Mr0grog: > `config/boot.rb` loaded within the default `bin/setup` takes care of this i believe:

`bundle exec` does actually [set several things about the ruby interpreter](https://github.com/bundler/bundler/blob/master/lib/bundler/shared_helpers.rb#L88-L93) before doing `require 'bundler/setup'` and that simply requiring `bundler/setup` can’t do. But yeah, *mostly* it’s paths, and you don’t need those if you are executing a specific file (`bin/setup`) instead of a command (`setup`), so it doesn’t matter 99% of the time. I guess it’s fine to not do it; most people won’t have a setup that causes a problem without it.

> Example of another big project that just uses `bin/setup`

I didn’t say we shouldn’t use it! BUT, that setup is also monumentally more complex than the 4 commands you are skipping by using the setup file here. They have a really good reason for scripting it. This project is stupidly tiny and simple.

Mr0grog:
**NOTE: This will need some cleanup if we want it, since #294 did not get merged.**

Issue #274, opened 2018-04-05
Re-implement `?include_versions` as a json:api-style relationship


Mr0grog:
In #264, we’re getting rid of `?include_versions` on `/pages` and the auto-included versions on `/pages/{id}`. It’s helpful functionality, though, so we’d like to bring it back as a [*relationship*](http://jsonapi.org/format/#document-resource-object-relationships), which allows us to include metadata and links so it can work as a chunked collection (because we can no longer afford to ever include *all* versions.

We should use the `?include_versions` query argument on *both* `/pages` and `/pages/{id}` (we previously only supported it on the collection, and treated the single object as if it was always there).

Mr0grog:
Tied in with #273 and should share its fate.

PR #310, opened 2018-05-08
Disable returning versions from /pages and /pages/{page_id}

### LINKED ISSUE #264: Remove `include_versions` from `/pages` and versions list from `/pages/{id}`
We have a nice, clever `?include_versions` querystring argument for the `/pages` route that causes each page object in the response to include a list of all its versions. We kind of knew that might be a problem down the road as we accumulated lots of versions, but with some pages now importing a lot of historical versions from Internet Archive, it’s going to be untenable very soon.

We should probably get rid of this option.

Alternatively, we could something like the N latest versions, where N is something like 10. If we did this, it might make sense to go further down the json:api route and classify them as relationships:

```js
{
  "links": {...},
  "data": [
    {
      "uuid": "1574d802-fd13-4699-881d-f0f8a7d45a79",
      "url": "https://www.fws.gov/refuges",
      "title": "National Wildlife Refuge System",
      "agency": "DOI",
      "site": "DOI - fws.gov",
      "created_at": "2017-05-09T05:26:41Z",
      "updated_at": "2018-03-28T05:31:57Z",
      "url_key": "gov,fws)/refuges",
      "maintainers": [...],
      "tags": [...]
      "relationships": {
        "versions": {
          "links": {
            "self": "https://api.monitoring.envirodatagov.org/api/v0/pages/1574d802-fd13-4699-881d-f0f8a7d45a79/versions",
            "first": "...",
            "last": "...",
            "next": "...",
            "previous": null
          },
          "data": [
            { /* version record */ },
            { /* version record */ },
            { /* version record */ }
          ]
        }
      }
    }
  ]
}
```

(But then maybe we ought to go back and fully implement json:api, and 💩 to me for not doing it right the first time around.)

**The same goes for the `/pages/{page ID}` route,** which always includes a `versions` property. It should also be removed or follow the same rules as above. (Maybe we could replace it with a `latest` and `earliest` property?)

/cc @danielballan @lightandluck @weatherpattern

lightandluck:
Fixes #264 

This is my first attempt at writing code for -db, so I'm probably not doing this idiomatically. I went the very conservative route of mostly commenting out code and skipping tests because we'll re-implement this functionality as described in #274. We'll need edgi-govdata-archiving/web-monitoring-ui#218 merged and deployed before deploying this.

Good news, I learned about implicit behavior in Rails and got to dig around in this repo!

### DIFF
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -18,7 +18,7 @@ class Api::V0::PagesController < Api::V0::ApiController
 
     result_data =
       if !should_allow_versions && should_include_versions
-        raise Api::NotImplementedError, 'The ?include_versions query argument has been disabled temporarily.'
+        raise Api::NotImplementedError, 'The ?include_versions query argument has been disabled.'
       elsif should_allow_versions && should_include_versions
         results = query
           .where(uuid: page_ids)
@@ -83,7 +83,7 @@ class Api::V0::PagesController < Api::V0::ApiController
     if should_allow_versions
       !should_include_versions && boolean_param(:include_latest)
     else
-      boolean_param :include_latest
+      boolean_param(:include_latest)
     end
   end
 
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -17,14 +17,7 @@ class Api::V0::PagesController < Api::V0::ApiController
     page_ids = id_query.pluck(:uuid, *order_attributes).collect {|data| data[0]}
 
     result_data =
-      if should_include_versions
-        results = query
-          .where(uuid: page_ids)
-          .includes(:versions)
-          .where(versions: { different: true })
-          .order('versions.capture_time DESC')
-        lightweight_query(results, &method(:format_page_json))
-      elsif should_include_latest
+      if should_include_latest
         # Filters from the original query should *not* affect what's "latest",
         # (though they do affect which *pages* get included) so we build a
         # whole new query from scratch here.
@@ -53,6 +46,9 @@ class Api::V0::PagesController < Api::V0::ApiController
   end
 
   def show
+    if should_include_versions
+      raise Api::NotImplementedError, 'The ?include_versions query argument has been disabled temporarily.'
+    end
     page = Page.find(params[:id])
     data = page.as_json(include: [:maintainers, :tags])
     data['versions'] = page.versions.where(different: true).as_json
@@ -70,7 +66,7 @@ class Api::V0::PagesController < Api::V0::ApiController
   end
 
   def should_include_latest
-    !should_include_versions && boolean_param(:include_latest)
+    boolean_param :include_latest
   end
 
   def page_collection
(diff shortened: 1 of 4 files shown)


app/controllers/api/v0/pages_controller.rb
```diff
@@ -17,14 +17,7 @@ def index
...
     result_data =
-      if should_include_versions
-        results = query
-          .where(uuid: page_ids)
-          .includes(:versions)
-          .where(versions: { different: true })
-          .order('versions.capture_time DESC')
-        lightweight_query(results, &method(:format_page_json))
-      elsif should_include_latest
+      if should_include_latest
```
Mr0grog: You need to keep the `if should_include_versions` branch here so you can throw an error!

Additionally, I think we need to keep the code in place and determine whether to throw an error or run it like we used to based on configuration. (Maybe `ALLOW_VERSIONS_IN_PAGE_RESPONSES`?) This is a pretty dangerous change, and if we find out that some other important service is relying on it in a way we didn’t know about, we need to be able to quickly turn it back on.

app/controllers/api/v0/pages_controller.rb
```diff
@@ -53,9 +46,16 @@ def index
   end
 
   def show
+    # NOTE: This check can be removed once this issue is resolved.
+    # https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/274
+    if should_include_versions
+      raise Api::NotImplementedError, 'The ?include_versions query argument has been disabled temporarily.'
+    end
```
Mr0grog: `should_include_versions` was *never* supported here, so I don’t think it makes sense to suddenly start throwing an error.

app/controllers/api/v0/pages_controller.rb
```diff
@@ -53,9 +46,16 @@ def index
...
     page = Page.find(params[:id])
     data = page.as_json(include: [:maintainers, :tags])
-    data['versions'] = page.versions.where(different: true).as_json
+    # NOTE: We'll re-attach versions to /pages/{page_id} when this issue is resolved.
+    # https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/274
+    # data['versions'] = page.versions.where(different: true).as_json
```
Mr0grog: Same issue here about keeping the functionality and gating it based on configuration.

app/controllers/api/v0/pages_controller.rb
```diff
@@ -70,7 +70,7 @@ def should_include_versions
   end
 
   def should_include_latest
-    !should_include_versions && boolean_param(:include_latest)
+    boolean_param :include_latest
```
Mr0grog: Same bit here about managing this functionality via configuration.

app/controllers/api/v0/pages_controller.rb
lightandluck: Oops! I totally mixed up where `?include_versions` was used. I was thinking it was on /pages/{page_id} instead of /pages for some reason. Fixed.

Mr0grog:
Ugh, so sorry this got left sitting forever. I think we all lost track of it through the mess of people traveling and being unable to work on this project much for the last couple months.

I think everything in other projects that needed to be fixed first has been merged and deployed, so this is good to go. Merging!

danielballan:
:+1:

lightandluck:
👍 🚀 🎉  😃

PR #349, opened 2018-07-28
BIGGER BETTER SEED DATA!!!!

### LINKED ISSUE #250: Improve the seed data
It doesn’t necessarily need to be bigger (though that would be nice), but it certainly needs to have more versions for each page. As it is, the experience of using a local DB with the seed data is kinda poor :\

What we should do here is pull down a selection of, say, 50 pages and all their versions from the staging or production DBs (but not include versions more recent than, say, mid-January 2018), format that as an import, and replace the existing `db/seed_import.json` with it.

Mr0grog:
It's pretty giant, but all but two entries have > 1 versions and there are 150 pages.

Fixes #250.


danielballan:
GOOD WORK!

PR #350, opened 2018-07-30
Add instructions for generating new seed data

Mr0grog:
I've been meaning to generate better seed data for a while (where most pages have multiple versions and where there are a fair number of pages), and finally did so for DCFemTech's hackathon this past weekend. I figured I'd add the script I wrote for it here so it's easy to run again if we need to (maybe to generate a different size import or just to create newer data).


Mr0grog:
Circle appears to be starting and then immediately shutting down Postgres before any of the tests run. Not sure what's going on — nothing’s changed in our configuration.

danielballan:
Nice!

danielballan:
xref https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/204

Issue #358, opened 2018-08-15
Allow for arbitrary hostname:port, md5 authentication in development and test environments


jsnshrmn:
Discussion about this started in https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/294
Currently, I'm only able to run this project in development and test environments if postgreSQL trusts the connection rather than challenging for a password.

## Expected Behavior
I'd expect to be able to configure some environment variables that allow me to develop this project against a password-protected database running on an arbitrary host:port.

## Current Behavior
Connections for development and test environments always unauthenticated requests to localhost on the default port.

## Possible Solution
The most straightforward way to do this is to parse DATABASE_URL and set some defaults based on that. PR forthcoming.

## Steps to Reproduce (for bugs)
1. set RAILS_ENV="development"
2. set pg_hba.conf to md5 auth instead of trust.
3. set posgresql server to listen on arbitrary port
4. rake db:*

## Context
I'm trying to get a closer approximation to production configuration in my development environment, so that I can gain more operational understanding of this app and automate more of the deployment process with confidence that we won't fall over or get hacked.

## Your Environment
CentOS 7
PostgreSQL 9.6

Mr0grog:
OK, we have a variety of different takes on solving DB config better in #294, #359, and #363.

It feels like there are a few different understandings and needs going on here; I’ve tried to pull them all together in #363, but need feedback to know if that’s clear and solves everyone’s different views here.

Mr0grog:
I’m considering the issues here resolved by #363.

PR #359, opened 2018-08-15
allow DB configuration in development and test environments

jsnshrmn:
See https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/358

This is neither super-pretty nor DRY, but it allows for more flexibility in the provisioning of dev and test environments. I'm leveraging the fact that the "yaml" file is really an erb yaml file. I'm templating in the  values inside of conditionals that should limit behavior changes to just the intended cases:
 when RAILS_ENV is development or test, and the user has bothered to specify DATABASE_URL.

I verified that this does what I need it to do when DATABASE_URL is set, and that it breaks in the same way it used to as described in https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/358 when RAILS_ENV is development or test and DATABASE_URL is not set.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -20,7 +20,10 @@ default: &default
   # For details on connection pooling, see rails configuration guide
   # http://guides.rubyonrails.org/configuring.html#database-pooling
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  <% if ENV['DATABASE_URL'] %><% db_uri = URI.parse(ENV['DATABASE_URL']) %>
+  # Try to set default connection properties based on DATABASE_URL.
+  # Otherwise, default to trusted localhost connections on the default port.
+  <% if ENV['DATABASE_URL'] %>
+  <% db_uri = URI.parse(ENV['DATABASE_URL']) %>
   host: <%= db_uri.host || 'localhost' %>
   port: <%= db_uri.port || '!!null' %>
   username: <%= db_uri.user || '!!null' %>
@@ -29,7 +32,15 @@ default: &default
 
 development:
   <<: *default
+  # Try to set the database to the value specified in DATABASE_URL if we're
+  # actually in the expected environment. Otherwise, default to
+  # web-monitoring-db_development
+  <% if ENV['RAILS_ENV'] == 'development' && ENV['DATABASE_URL'] %>
+  <% db_uri = URI.parse(ENV['DATABASE_URL']) %>
+  database: <%= db_uri.path || web-monitoring-db_development %>
+  <% else %>
   database: web-monitoring-db_development
+  <% end %>
 
   # The specified database role being used to connect to postgres.
   # To create additional roles in postgres see `$ createuser --help`.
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -20,11 +20,27 @@ default: &default
...
+  username: <%= db_uri.user || '!!null' %>
+  password: <%= db_uri.password || '!!null' %>
+  <% end %>
 
 development:
   <<: *default
+  # Try to set the database to the value specified in DATABASE_URL if we're
+  # actually in the expected environment. Otherwise, default to
+  # web-monitoring-db_development
+  <% if ENV['RAILS_ENV'] == 'development' && ENV['DATABASE_URL'] %>
```
Mr0grog: Checking `ENV['RAILS_ENV']` is not a good way to check the environment for two reasons:

- There are several ways an environment can get set in Rails, including the `RAILS_ENV` OR `RACK_ENV` environment variables, so this doesn’t adequately check whether we are actually in the development environment.

- Various method calls and situations within Rails can change the environment and I’m not totally certain they reliably update `ENV`.

To check the environment in Rails, you should use `Rails.env`, like so:

```rb
if Rails.env.development?
  # do whatever
end
```

config/database.yml
```diff
@@ -58,7 +74,15 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *default
+  # Try to set the database to the value specified in DATABASE_URL if we're
+  # actually in the expected environment. Otherwise, default to
+  # web-monitoring-db_test
+  <% if ENV['RAILS_ENV'] == 'test' && ENV['DATABASE_URL'] %>
```
Mr0grog: Same note about environments as above.

config/database.yml
```diff
@@ -58,7 +74,15 @@ development:
...
+  <% db_uri = URI.parse(ENV['DATABASE_URL']) %>
+  database: <%= db_uri.path || web-monitoring-db_test %>
+  <% else %>
   database: web-monitoring-db_test
+  <% end %>
```
Mr0grog: This doesn’t work if I’m in a dev environment and I’m running a command like `rake db:schema:load` and my test DB has a different user from my dev DB or it has a different name. (When the rake command flips to repeat in the test environment, it’s picking up this config, but the users from my `DATABASE_URL`, and it gets messy.)

jsnshrmn:
Thank you for the thoughtful review! I feel like I'm picking a fight with the way railsconfig and rake db commands interact with this PR. Clearly setting defaults is not quite the right thing to do: doing wrong stuff with a potentially remote database is mostly going to be worse than doing wrong stuff against a definitely local database. I can see that I'm being a bad citizen on that front.

jsnshrmn:
Closing this out with wontfix, as I've verified that https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/363 solves https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/358, and is generally a much tidier and less problematic solution that what's on offer here.

PR #384, opened 2018-09-12
Add new S3 buckets to ALLOWED_ARCHIVE_HOSTS

Mr0grog:
We're migrating S3 data between AWS accounts, which also means making new buckets. Instead of `edgi-versionista-archive`, we now have `edgi-wm-versionista`. We're still allowing the old buckets, though, because we haven't actually moved all the data yet.

This is a subtask of edgi-govdata-archiving/web-monitoring#101.


Mr0grog:
This is just an already-agreed-upon update to example configuration, so I’m not going to worry about review here.

PR #397, opened 2018-09-26
Add data migration for new S3 buckets

Mr0grog:
This adds the migration code as a rake task. We've talked before about issues with extremely large data migrations, so this is my first take at moving them out of the normal migrations path by making each one a rake task. I'm thinking we'll namespace them all under `:data`.

This is part of edgi-govdata-archiving/web-monitoring#101.


Mr0grog:
Merging this so we can move ahead with migrating S3 data.

PR #398, opened 2018-09-26
Update seed data to use new S3 buckets

Mr0grog:
Basically just a big find-and-replace job. It also updates the `.env.example` file because I realized our non-versionista bucket was not listed. You wouldn’t want this in production, but you probably *would* want it if you’re just getting set up and don’t want to download all the raw seed data from S3 :)

This is a part of edgi-govdata-archiving/web-monitoring#101.


Mr0grog:
I’m going to go ahead and merge this since no sane human is going to want to read this diff.

PR #401, opened 2018-10-03
Stop archiving if the hash is known and archived

### LINKED ISSUE #377: Archiver should not do anything for content we already have by hash
When the archiver stores content, it first downloads the content, then verifies the hash, then checks whether it’s already stored, and then stores it: 
https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/6ed10a27a0749001482b3ed5e8623fca22c4f6f9/lib/archiver/archiver.rb#L34-L57

However, if the archiver was provided with a hash, we should just check whether anything is stored in that hash and stop immediately if so. There’s no need to do any downloading or verification at that point. This can also save us a lot of unnecessary writes to S3, which are expensive.

Mr0grog:
Since we store files by their hash, we know that a hash that is already in the store should be good, so we can stop early; there's [generally] no reason to download the content at all. If you *really* want to verify that the content at the URL to archive matches the already known hash, you can set the `force: true` argument.

This also updates the S3 storage layer to actually check whether the object is in the bucket (as opposed to just whether the bucket is correct). I think I had intended to do this before, but never got around to it! :(

Finally, the order of checks in `Archiver.already_archived?()` is reversed so that the cheapest check is first (there's no network hit for determining whether the URL is at an allowable, external archive host).

Fixes #377.

### DIFF
--- a/lib/archiver/archiver.rb
+++ b/lib/archiver/archiver.rb
@@ -31,7 +31,13 @@ module Archiver
 
   # Primary API ----------
 
-  def self.archive(url, expected_hash: nil)
+  def self.archive(url, expected_hash: nil, force: false)
+    # If the hash is already in the store, there's no reason to load & verify.
+    if expected_hash && !force
+      hash_url = store.url_for_file(expected_hash)
+      return { url: hash_url, hash: expected_hash } if store.contains_url?(hash_url)
+    end
+
     response = retry_request do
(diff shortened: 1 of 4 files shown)


Mr0grog:
Alright, I’m going to go ahead and merge this in shortly since it’s been sitting for quite a while.

PR #402, opened 2018-10-03
Ensure AWS region is set for internal/private storage

Mr0grog:
We correctly set the region for public archival storage, but not for internal, private storage of working files. The old storage bucket was in `us-east-1` (the default), so this was easy to miss until edgi-govdata-archiving/web-monitoring#101 was solved.


Mr0grog:
(Note the current state of the app still works fine; not having the region set correctly for S3 just slows things down a bit because we get redirects *and* every redirect means the request also has to be signed again.)

PR #403, opened 2018-10-03
Make jobs run a little safer on Heroku

Mr0grog:
While trying to get Wayback imports working reliably, I’ve noticed that we frequently wind up with job progress stuck hanging for hours and hours and hours. Obviously we need to give up at some point on the ETL script side, but it’s not like any import jobs are actually taking that long.

Instead, what’s really happening is that Heroku has some Resque-incompatible behavior when it randomly kills dynos every so often. Specifically, it sends SIGTERM to both the Resque process *and to the process’s children,* but Resque’s children aren’t prepared for that — the children simply die and the parent never tracks the fact that the job didn’t finish and so never re-enqueues it to run later. The `resque-heroku-signals`  gem *should* solve that issue.

Additionally, since the Wayback import jobs are so big, it would be nice to have some sense of progress in the API and the database. This now persists updates to the `Import` record every 5 seconds (so as to not cause a lot of overhead that slows down processing an import).


Mr0grog:
Alright, I’m going to go ahead and merge this, since nobody has reviewed.

Issue #404, opened 2018-10-04
Non-Versionista imports need equivalent source & text diff hashes


Mr0grog:
We’ve long planned to add [automated analysis for versions](https://github.com/edgi-govdata-archiving/web-monitoring/issues/62) whenever they get imported, but I had failed to mentally connect the fact that not having done this means our Wayback imports don’t have source & text diff hashes to the fact that analysts heavily rely on those for managing their work. I think this means having some equivalent information for Wayback imports is a high priority for making our Wayback data workable.

There are a few possible solutions here:

1. Do option (2) from edgi-govdata-archiving/web-monitoring#62, which is the right long-term solution here.
    - 👍 It’s generally the right future architecture we’ve agreed on
    - 👎 It requires invention of a whole infrastructure around queuing. We can’t use the same job queue we are using in Rails because it’s not compatible with our Python code. We can’t *just* use SQS because we are committed to a) making the platform workable locally, offline, and b) being minimally dependent on vendor-specific solutions, so we’d need an abstraction layer than can use SQS or something else local.
    - 👎 It requires coordinated changes across 3 projects: DB has to put things on the queue; Processing has to add queue reading and an analysis module; Scraper has to learn to load annotations for every version and use those in its weekly sheets output.

2. Rig up a new job type in DB that loads the appropriate diffs from the diffing server, hashes them, and adds them to either the `source_metadata` (a small lie on the DB’s part to ease compatibility) or makes an annotation with them (requires extra work on scraper’s part).
    - 👍 it’s quick and expedient
    - 👍 it relies on existing infrastructure on only requires changing 1 (or maybe 2) projects
    - 👍 the diffs themselves get pre-cached because DB loaded them! 😀
    - 👎 it adds a whole new body of logic in Ruby when we are trying to switch to Python (#119) and puts that logic in the DB instead of processing (although there’s probably still some debate or discussion worth having there).
    - 👎 It’s an interim solution that is definitely not the right one, and it’s easy for that to get conceptually wedged into place and become a roadblock for getting the right solution.

3. Add an analysis endpoint to the diffing server (or a new, separate server) in the processing project that receives a POST of the DB records for two versions + page (or maybe three versions: a, b, base) and is responsible for POSTing the resulting annotation back to the DB. We’d add a new job type in DB that sends that initial POST to the analysis endpoint.
    - 👍 Does not require any new queuing infrastructure as in option (1) (we’re leaning on DB to manage queuing)
[...]

PR #406, opened 2018-10-08
Auto-analyze newly imported versions

Mr0grog:
This is a quick implementation of approach 2 from #404. If we went with approach 3, we’d basically just be porting `AnalyzeChangeJob#analyze_change` over to a new server & endpoint in the processing project, and the method here would just call it.


Mr0grog:
Thanks, @jsnshrmn! I’m going to wait until late tonight to give @danielballan a chance to at least weigh in on this conceptually before pushing it in.

Mr0grog:
Side note: I’ve created users in both environments and added the username & password to our secrets files so that’s ready to go regardless of what we choose to do here. That data will be needed no matter what since we’ll always need to attribute the annotation to someone.

Mr0grog:
> Just wondering if that could be an accidentally-committed file.

Yep.

Mr0grog:
I am also going to change this to scale the source/text component of the priority by the percentage change.

PR #408, opened 2018-10-09
 Add `?include_change_from_*` to version endpoints

### LINKED ISSUE #404: Non-Versionista imports need equivalent source & text diff hashes

Mr0grog:
This adds to additional query params to endpoints that return a version or a list of versions:

- `?include_change_from_previous`
- `?include_change_from_earliest`

That include a field with the given name whose value the the change object (if any) representing the the change relative to the requested version. This is implemented *really* inefficiently (so it's very good that you have to ask for it), but I couldn't think of any way to do this better without adding a lot of denormalization. Regardless, I think it's still faster to have the API include this feature than to have the scraper make more API calls when building analyst task sheets.

(One alternative I’m imagining is querying for all Change objects related to captures in a given timeframe via a non-existent `/changes` endpoint and then manually correlating those with a list of Versions. But that’s a *lot* more work for an API consumer.)

This also depends on #406. I was hoping to avoid that, but big API changes here mean it intersects with that PR. Fixes #404.

Note: check just this commit to see the changes here that aren’t part of #406: https://github.com/edgi-govdata-archiving/web-monitoring-db/pull/408/commits/c27146ff82727df1474c01c57a6ab6ba8c19ff98

### DIFF
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -9,7 +9,7 @@ class Api::V0::VersionsController < Api::V0::ApiController
     render json: {
       links: paging[:links],
       meta: { total_results: paging[:total_items] },
-      data: versions
+      data: versions.collect {|version| serialize_version(version)}
     }
   end
 
@@ -21,7 +21,7 @@ class Api::V0::VersionsController < Api::V0::ApiController
         previous: @version.previous && api_v0_version_url(@version.previous),
         next: @version.next && api_v0_version_url(@version.next)
       },
-      data: @version
+      data: serialize_version(@version)
     }
   end
 
@@ -109,5 +109,14 @@ class Api::V0::VersionsController < Api::V0::ApiController
     collection = where_in_range_param(collection, :capture_time, &method(:parse_date!))
 
     sort_using_params(collection)
+  end
+
+  def serialize_version(version, options = {})
+    methods = options[:methods] || []
+    methods << :change_from_previous if boolean_param(:include_change_from_previous)
+    methods << :change_from_earliest if boolean_param(:include_change_from_earliest)
+    options[:methods] = methods
+
+    version.as_json(options)
   end
 end
--- a/app/jobs/analyze_change_job.rb
+++ b/app/jobs/analyze_change_job.rb
@@ -7,7 +7,7 @@ class AnalyzeChangeJob < ApplicationJob
     change = if from_version
       Change.between(from: from_version, to: to_version)
     else
-      to_version.change_from_previous
+      to_version.ensure_change_from_previous
     end
 
     return unless is_analyzable(change)
@@ -15,9 +15,8 @@ class AnalyzeChangeJob < ApplicationJob
     analyze_change(change)
 
     if compare_earliest
-      earliest = to_version.page.versions.reorder(capture_time: :asc).first
-      change_from_earliest = Change.between(from: earliest, to: to_version)
-      analyze_change(change_from_earliest) if is_analyzable(change_from_earliest)
+      earliest_change = to_version.ensure_change_from_earliest
+      analyze_change(earliest_change) if is_analyzable(earliest_change)
     end
   end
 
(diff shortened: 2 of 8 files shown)


Mr0grog:
> One alternative I’m imagining

Actually, that wouldn’t even work, since you’d still need to know whether a given change is the change from the *previous* (or earliest) version. So never mind on that alternative. *sigh.*

Mr0grog:
Rebased after merging #406 so this is easier to read.

Mr0grog:
OK, based on discussions at the meeting today, it seemed like folks were generally happy with merging this. I’m going to do that so I can show what the data here looks like in practice at tonight’s analyst meeting, but please comment or review here if you want to and have thoughts about the implementation that I should change.

PR #409, opened 2018-10-09
Add a basic /healthcheck endpoint with test.

### LINKED ISSUE #385: Add health check endpoint
To support various tools (and in particular, Kubernetes), we should add an HTTP health check endpoint to the server. It doesn’t technically matter what it returns (but we should be consistent and return some JSON) so long as it is able to quickly return a 200 status code.

Other tools and use this to determine whether the server is healthy (and, if they have the rights, optionally kill and restart it).

I’m thinking it should be at `GET /healthcheck`.

See also edgi-govdata-archiving/web-monitoring-processing#154.

danielballan:
Closes #385

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -1,5 +1,9 @@
 class HealthcheckController < ApplicationController
   def index
+    # Ensure database is available
+    ActiveRecord::Base.connection.exec_query('SELECT 2 + 2;')
+    # Ensure job queue is available
+    Resque.size(:not_a_real_queue_but_thats_ok)
     render json: {}
   end
 end
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -1,9 +1,21 @@
 class HealthcheckController < ApplicationController
   def index
-    # Ensure database is available
-    ActiveRecord::Base.connection.exec_query('SELECT 2 + 2;')
-    # Ensure job queue is available
-    Resque.size(:not_a_real_queue_but_thats_ok)
-    render json: {}
+    # Succeed if the app is running, but provide additional
+    # info if related components have gone down.
+    db = 'ok'
+    begin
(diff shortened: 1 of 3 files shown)


danielballan:
Please review the test with appropriate skepticism, as I am straight cargo-culting from existing test code.

app/controllers/healthcheck_controller.rb
```diff
@@ -0,0 +1,5 @@
+class HealthcheckController < ApplicationController
+  def index
+    render json: {}
```
Mr0grog: Not sure whether you think it makes sense to also check that the DB and queues are operational. If so, you could add this before the `render` call:

```rb
# Ensure database is available
ActiveRecord::Base.connection.exec_query('SELECT 2 + 2;')
# Ensure job queue is available
Resque.size(:not_a_real_queue_but_thats_ok)
```

Both will raise exceptions if they can’t connect.

test/controllers/healthcheck_controller_test.rb
```diff
@@ -0,0 +1,12 @@
+require 'test_helper'
+require 'minitest/mock'
+
+class Api::V0::DiffControllerTest < ActionDispatch::IntegrationTest
+  include Devise::Test::IntegrationHelpers
```
Mr0grog: Minor note: you shouldn’t need this since you don’t need to do auth for this endpoint (`Devise` is the authentication library we use).

app/controllers/healthcheck_controller.rb
danielballan: The `Rescue...` line errors out:

```
Error:
Api::V0::DiffControllerTest#test_healthcheck_responds_with_success:
Errno::EADDRNOTAVAIL: Cannot assign requested address - connect(2) for [::1]:6379
    app/controllers/healthcheck_controller.rb:6:in `index'
    test/controllers/healthcheck_controller_test.rb:7:in `block in <class:DiffControllerTest>'
```

Is that quite right?

PR #410, opened 2018-10-10
Automatically publish docker images from release branch

Mr0grog:
We previously used CircleCI’s “approval” feature for managing the publishing of docker images from the `master` branch. Unfortunately, experience has shown us that what sounds like a promising feature is actually poorly executed enough that it causes confusion and wasted build time on CircleCI. Because nobody gets notified that a job is awaiting approval and jobs can't expire (see *many* CircleCI support tickets requesting this from plenty of users), workflow approvals often get forgotten and are then stuck hanging until someone comes along to clean them out.

Instead, we're adding a `release` branch. Any pushes to this branch will get automatically published to Docker Hub, no approval needed. Management is now handled by someone manually adding a commit from `master` onto `release`.

I’m going to try publishing #406 using this mechanism instead of the approval system to test it out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,9 @@ workflows:
       - build:
           filters:
             branches:
-              ignore: master
+              ignore:
+                - master
+                - release
 
   build-and-publish:
     jobs:
@@ -74,11 +76,7 @@ workflows:
           filters:
             branches:
               only:
-                - master
-      - hold:
-          type: approval
-          requires:
-            - build
+                - release
       - publish_docker:
           requires:
-            - hold
+            - build


.circleci/config.yml
```diff
@@ -66,19 +66,17 @@ workflows:
       - build:
           filters:
             branches:
-              ignore: master
+              ignore:
+                - master
```
danielballan: Do we want to keep ignoring `master` here, or just ignore `release`?

jsnshrmn:
Also, I ❤️ this branch name.

Mr0grog:
Neat! Can you make an issue with those links and info? For now, I don’t think it makes sense to hold this up. Implementing it doesn’t seem straightforward because part of the goal here was to *not* publish on every commit to master.

jsnshrmn:
Oh sure, I wasn't suggesting that we automate it right now, I was just pointing to the merge strategy as a handy way to track deployments without having to resolve conflicts or rewrite history. I was just imagining you could do that manually. I will go ahead and write up an issue for automation.

Mr0grog:
Ah, I see! I’ll have to start doing them this way. Having those nice clear commits that identify themselves as releases/deploys is good.

.circleci/config.yml
Mr0grog: Ack, somehow I missed your comment here. I guess I figured we didn’t need to test master if we already do every other branch before merging. But I guess it’s not a bad idea in case someone lands something directly on master for some reason. I’ll change it.

PR #412, opened 2018-10-12
Update `copy_page` rake task to match current API

Mr0grog:
Since we wrote this rake task, we've made changes to the API (added maintainers and tags, removed versions from the page endpoint). We haven't updated this task to match, though, and so it always fails. This makes some sorely needed updates.

I’ve had a rough version of this sitting in my working copy for a while, and it’s high time I cleaned it up so this task is usable for others.


Mr0grog:
> I could see how this could be handy for instance, if someone wanted to pull down data (with matching IDs) from staging into a local dev environment to troubleshoot an issue or just test against real data.

Yep, that’s exactly what it was originally written for 😄 (#183)

Issue #411, opened 2018-10-11
Clean up ugly hotfix from diff-related cluster collapse on 2018-10-10


Mr0grog:
I implemented a pretty ugly emergency fix today for some cluster issues we were having: https://github.com/edgi-govdata-archiving/web-monitoring-db/commit/6ea87de05d52d823f989e531dc1a600e25edab25

When trying to diff two versions, we check their content-type to make sure it's reasonable to even try. Unfortunately, we don’t have content-type metadata for really old Versionista versions, so we fall back to assuming it might work and giving the diff a try. The hotfix addresses this by checking the file extension (if present) in the URL, too. That probably needs to stay as a necessary step in the long term, so I need to come back and clean up the logic to make it more reasonable.

---

(To be clear, the diffing server needs to become more robust, too. See edgi-govdata-archiving/web-monitoring-processing#154 and edgi-govdata-archiving/web-monitoring-processing#287. That shouldn’t prevent us from behaving better on this side of things, too, though.)

PR #415, opened 2018-10-14
Clean up and loosen auto-analysis hotfixes

Mr0grog:
This changes several things around our auto-analysis code to firm it up a bit:

1. Cleans up the overall hotfix-related logic so it’s less janky (#411)
2. Loosens the restrictions put in place by the hotfix. Now that we’ve merged edgi-govdata-archiving/web-monitoring-processing#289, we don’t need the strict safety and we can attempt more diffs. **However, this also adds an `ANALYSIS_REQUIRE_MEDIA_TYPE` env var you can set to get the strict, hotfix-style behavior back an emergency.**
3. Splits jobs into two separate queues so they are easy to move to separate processes/images/pods/whatever later. There is now an `import` queue and an `analysis` queue. By default, the Import Worker image runs both, but you can change that with the `QUEUES` environment variable.
4. Adds a rake task for queuing up new analyses.

I’m going to go ahead and merge this in the morning if I don’t see any glaring errors after sleeping on it. That’s not our usual policy, but I want to do our best to have something that works for the analysts on Monday night, and we’ll need some time to run a week’s work of analyses.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -68,3 +68,7 @@ ALLOW_VERSIONS_IN_PAGE_RESPONSES='false'
 # Set this in production to look up the user associated with automatic
 # annotations. Note a user with this e-mail must exist.
 # AUTO_ANNOTATION_USER='someone@example.com'
+
+# Set to 'true' to require versions to have valid mime/content/media type fields
+# in order to be automatically analyzed after importing.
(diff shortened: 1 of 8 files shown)


Mr0grog:
This is way outside standard operating procedure, but I’m going ahead and merging this so we can get analysis going for analysts for their Monday night spreadsheet run.

PR #418, opened 2018-10-16
Exempt the healthcheck endpoint from force_ssl.

danielballan:
This should allow Kubernetes liveness/readiness probes to make HTTP requests.

per documentation: https://edgeapi.rubyonrails.org/classes/ActionDispatch/SSL.html

via this SO post: https://stackoverflow.com/a/38572725/1221924

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -41,6 +41,12 @@ Rails.application.configure do
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
+  # ...but exempt the healthcheck:
+  config.ssl_options = {
+    redirect: {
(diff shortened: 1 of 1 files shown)


jsnshrmn:
So, I deployed this change to staging and replaced the tcp port check with the http get check.  The containers didn't come up, logs showed the following:
```
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
2018-10-17 19:33:24 +0000: HTTP parse error, malformed request (): #<Puma::HttpParserError: Invalid HTTP format, parsing fails.>
---
2018-10-17 19:33:26 +0000: HTTP parse error, malformed request (): #<Puma::HttpParserError: Invalid HTTP format, parsing fails.>
```
I went ahead and rolled back.

PR #421, opened 2018-10-21
Add retry/backoff logic to SimpleDiff

### LINKED ISSUE #414: SimpleDiff should have retry logic
Now that we are auto-analyzing every change, we are talking to the differ almost all the time. That means that, when we are deploying updates, we are likely to get the occasional failed connection (even a rolling update might cut a diff off in the middle sometimes). `Differ::SimpleDiff#generate_diff` should use some simple retry + backoff logic so these kind of quick, spurious failures don’t totally trip it up.

`Archiver` does this already, so we can probably copy or extract that code: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/e7cd32e52cfb2a6eaef00486d0f017116193a0e8/lib/archiver/archiver.rb#L84-L96

Mr0grog:
Hopefully this is pretty straightforward. We should generally be expecting that any upstream HTTP service could have intermittent failures (general network problems, machine being restarted, etc). We weren’t doing that with `SimpleDiff`, but this adds it.

Fixes #414.

### DIFF
--- a/lib/archiver/archiver.rb
+++ b/lib/archiver/archiver.rb
@@ -87,7 +87,7 @@ module Archiver
       begin
         response = yield
         return response if attempt >= tries || (response.code != 503 && response.code != 504)
-      rescue HTTParty::ResponseError => error
+      rescue HTTParty::ResponseError, Timeout::Error => error
         raise error if attempt >= tries
       end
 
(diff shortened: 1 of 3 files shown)


Mr0grog:
I’m going ahead and merging this. It’s a critical solve for a persistent problem in production.

PR #434, opened 2018-11-22
Patch HTTParty to handle bad Wayback gzip headers

### LINKED ISSUE #433: Handle bad gzipped content from Wayback
See also edgi-govdata-archiving/web-monitoring-processing#309.

When requesting a memento that Wayback originally scraped a URL that responded with the `Content-Encoding: gzip` header, Wayback screws up the same header in the memento response, resulting in gzipped content over the wire that our (or any) HTTP library doesn’t automatically decode for us (because it gets `Content-Encoding: ` [empty value for the header]). Then we wind up storing, reading, and diffing gzipped bytes instead of the actual HTML source (or whatever the page was if not HTML).

The simple workaround here is to set the `Accept-Encoding: identity` header on the request.

On the other hand, if we want gzipping to reduce network time and costs, this seems to be a workable hack:

```ruby
module HTTParty
  class Response
    core_init = self.instance_method(:initialize)

    define_method(:initialize) do |request, response, parsed_block, options = {}|
      core_init.bind(self).call(request, response, parsed_block, options)

      if headers['content-encoding'] == ', gzip'
        headers['content-encoding'] = 'gzip'
        ungzip_body
      end
    end

    # Torn unceremoniously from an old version of HTTParty, when it was removed
    # (because Ruby's Net::HTTP library should already do this, except for the
    # header bug)
    # https://github.com/jnunemaker/httparty/commit/6f6bf6b726484eaf50e190769bbe14c9841a2c64
    def ungzip_body
      body_io = StringIO.new(body)
      self.body.replace Zlib::GzipReader.new(body_io).read
    end
  end
end
```

A slightly less dangerous fix might be to patch the `headers` and `body` methods, which are just accessors and so don’t take arguments and are less likely to change than the initializer.

Mr0grog:
In some situations, Wayback screws up the `Content-Encoding` header when sending gzipped data (see the comments in the patch here or the original issue for more details). They don't have any clear timeline on fixing the issue, so this hack may stay in indefinitely. If we ever switch HTTP libraries, we'll unfortunately have to figure out how to fix it or stop using gzip.

The alternative is to send `Accept-Encoding: identity` to disable gzipping for this content entirely. I’d like to avoid that since we are often pulling in a lot of data, but it is the safer fix. Happy to entertain objections on that front :)

Fixes #433.

### DIFF
--- a/config/initializers/httparty.rb
+++ b/config/initializers/httparty.rb
@@ -1,0 +1,50 @@
+# The Wayback Machine unfortunately has some bugginess around gzip encoding
+# that we are working around here.
+#
+# If you request a memento (e.g. a URL like:
+# `http://web.archive.org/web/20181118000000id_/http://epa.gov`) with an
+# `Accept-Encoding: gzip` header, Wayback will happily gzip the response body
+# and include a `Content-Encoding: gzip` header. However, if the
+# original snapshot was gzipped when captured by Wayback, it screws this up --
+# the memento response body will be gzipped, but it will include two encoding
+# headers: `content-encoding: ` (that's right, it has no value) and then later,
+# `Content-Encoding: gzip`. Pretty much any client in this case will obey the
+# first one and assume there is no special encoding, so we'll get gzip bytes
+# out our end instead of the uncompressed response data we wanted.
+#
+# This little hack amends HTTParty's Response class to look for this malformed
+# header situation, clean up the headers, and decompress the body.
+#
+# NOTE: you can also avoid this situation by never requesting gzipped responses
+# (i.e. not sending `Accept-Encoding` or sending `Accept-Encoding: identity`).
+# We are loading lots of data, though, and having it gzipped over the network
+# when possible is a huge boon we'd like to keep.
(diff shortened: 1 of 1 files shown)


Mr0grog:
I think I’m mainly looking for a gut-check here on whether it’s the right thing to do this or if we should just send `Accept-Encoding: identity` (i.e. turn off gzip support) with all requests to `archive.org`.

Mr0grog:
@jsnshrmn @danielballan I think I’m mainly looking for a gut-check from one of you on whether it’s the right thing to do this or if we should just send Accept-Encoding: identity (i.e. turn off gzip support) with all requests to archive.org.

Mr0grog:
And if folks don’t feel good about this, I’ll revert this processing one, too: https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/174/commits/0e896eed2f75347284cd7418a3c6b8d15da8c8e3

Mr0grog:
Sweet, I’ll merge and release! 🎉

PR #436, opened 2018-11-26
Ensure Version/Page titles are normalized to be a single line, remove surrounding spaces, etc.

### LINKED ISSUE #430: Version/Page `title` field should be one line
I just had to write a quick fix in our analyst sheet generation script to handle `title` fields with line breaks in them (edgi-govdata-archiving/web-monitoring-versionista-scraper#165). Obviously the ETL script for wayback data should be removing those before passing it to the DB in the first place (in practice, that’s the root source of the problem), but it seems reasonable that the DB should sanitize the `title` field when importing, too.

In general, we should treat the `title` field for Page and Version objects like a web browser treats it in HTML: remove surrounding whitespace and treat consecutive whitespace characters (including line breaks) inside the string as a single blank space.

The best place for this logic is probably a `before_save` hook on the Page and Version models.

Mr0grog:
Fixes #430 by adding a concern (a mixing for models) that normalizes the `title` attribute before saving. Both the `Page` and `Version` models use it. I’m thinking it might make sense to extend to tags and maintainers in the future, but that’s out of scope here.

Also adds a data migration to fix existing records with bad titles.

### DIFF
--- a/app/models/concerns/simple_title.rb
+++ b/app/models/concerns/simple_title.rb
@@ -1,0 +1,21 @@
+module SimpleTitle
+  extend ActiveSupport::Concern
+
+  included do
+    before_save :normalize_title
+  end
+
+  class_methods do
+    # A class using this concern can provide its own implementation of
+    # normalize_title_string(string) to customize the normalization.
+    def normalize_title_string(title)
+      title.strip.gsub(/\s+/, ' ')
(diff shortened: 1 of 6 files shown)


Mr0grog:
Gonna go ahead and merge at this point.

PR #437, opened 2018-11-27
Add route for raw version bodies.

### LINKED ISSUE #431: Add route for raw version bodies
This issue has come up a huge number of times, but it turns out we don’t have an issue tracking it. The `uri` field of a Version currently holds a URI indicating where the raw response body of the version is stored, *regardless of whether a client can access that URL*. That’s a real problem when the URL is not public — for example, when using the `LocalFile` file store, the URLs are `file://` URLs. In these cases, we need to provide URLs that are publicly accessible.

First, we should add a route that returns either:
- The raw response body
- A permanent redirect (`301`) to the storage URI if the it’s a public location

Some ideas for this route:
- `/api/v0/versions/{uuid}/raw`
- `/api/v0/versions/{uuid}/body`
- ?

Second, we should make sure that everywhere we return version records (e.g. `/versions`), we replace the `uri` field with the above URL. (If the `uri` is not publicly accessible?)

jsnshrmn:
I believe this fixes #431, though I implemented it differently than @Mr0grog suggested in that issue. This PR provides a couple of new routes and some changes to `FileStorage::LocalFile`:

### A route that serves up a raw response body from local storage.

- `/api/v0/raw/{hash}`
- `FileStorage::LocalFile.url_for_file` now returns this value instead of a `file://` URL
- file storage and archiver tests now check for these urls

I created this route so that LocalFile could mint usable URLs without knowing about UUIDs. All new locally stored versions are now accessible via the network.

### A shortcut route that redirects to the URI for a version:

- `/api/v0/versions/{uuid}/raw`
- Returns a 301 to whatever location is specified in that version's URI field.

Compared to the suggested implementation, I've reduced this route to a simple shortcut. Local files are the only case of inaccessible snapshots that I am addressing here, and this code assumes that since those are now accessible, all URIs are accessible. That could be wrong. Also, it is adding an extra request for every local file served. If that's a problem, this route could directly return the body as well, although then it would provide two canonical locations for identical content.

### DIFF
--- a/app/controllers/api/v0/raw_controller.rb
+++ b/app/controllers/api/v0/raw_controller.rb
@@ -2,9 +2,8 @@ class Api::V0::RawController < Api::V0::ApiController
 
   def show
     @version ||= Version.find_by(version_hash: params[:id])
-    # TODO: only return a local file if a bucket isn't set. Otherwise redirect to bucket
-    # TODO: don't hardcode path
-    path = '/app/tmp/storage/archive/' + @version.version_hash
+    storage_path = Rails.root.join 'tmp/storage/archive'
+    path = "#{storage_path}/#{@version.version_hash}"
     mime_type = @version.source_metadata['mime_type']
     send_file(path, type: mime_type, disposition: 'inline')
   end
--- a/app/controllers/api/v0/raw_controller.rb
+++ b/app/controllers/api/v0/raw_controller.rb
@@ -1,0 +1,11 @@
+class Api::V0::RawController < Api::V0::ApiController
+
+  def show
+    @version ||= Version.find_by(version_hash: params[:id])
+    # TODO: only return a local file if a bucket isn't set. Otherwise redirect to bucket
+    # TODO: don't hardcode path
+    path = '/app/tmp/storage/archive/' + @version.version_hash
+    mime_type = @version.source_metadata['mime_type']
+    send_file(path, type: mime_type, disposition: 'inline')
+  end
+end
--- a/app/controllers/api/v0/raw_controller.rb
+++ b/app/controllers/api/v0/raw_controller.rb
@@ -8,8 +8,4 @@ class Api::V0::RawController < Api::V0::ApiController
     send_file(path, type: mime_type, disposition: 'inline')
   end
 
-  # This is just a placeholder so the router can build a URL for file_storage.
-  def index
-    head 404
-  end
 end
--- a/app/controllers/api/v0/raw_controller.rb
+++ b/app/controllers/api/v0/raw_controller.rb
@@ -7,4 +7,9 @@ class Api::V0::RawController < Api::V0::ApiController
     mime_type = @version.source_metadata['mime_type']
     send_file(path, type: mime_type, disposition: 'inline')
   end
+
+  # This is just a placeholder so the router can build a URL for file_storage.
+  def index
+    head 404
+  end
 end
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -42,7 +42,6 @@ class Api::V0::VersionsController < Api::V0::ApiController
     elsif !Archiver.already_archived?(@version.uri) || !@version.version_hash
       result = Archiver.archive(@version.uri, expected_hash: @version.version_hash)
       @version.version_hash = result[:hash]
-      # TODO: set the uri to the raw path if result is a local file.
       @version.uri = result[:url]
     end
 
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -30,10 +30,8 @@ class Api::V0::VersionsController < Api::V0::ApiController
     # TODO: only return a local file if a bucket isn't set. Otherwise redirect to bucket
     # TODO: don't hardcode path
     path = '/app/tmp/storage/archive/' + @version.version_hash
-    # handy for testing since it renders html, but pretty mucks with the content.
-    #render file: '/app/tmp/storage/archive/' + @version.version_hash
-    # TODO: sniff mime type
-    send_file(path)
+    mime_type = @version.source_metadata['mime_type']
+    send_file(path, type: mime_type, disposition: 'inline')
   end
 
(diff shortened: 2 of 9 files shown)


app/controllers/api/v0/raw_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Api::V0::RawController < Api::V0::ApiController
+
+  def show
+    @version ||= Version.find_by(version_hash: params[:id])
+    storage_path = Rails.root.join 'tmp/storage/archive'
+    path = "#{storage_path}/#{@version.version_hash}"
+    mime_type = @version.source_metadata['mime_type']
+    send_file(path, type: mime_type, disposition: 'inline')
```
Mr0grog: So… unless I’m really missing something, this is straight-up not going to work in production or any environment that is not using the `LocalFile` Filestore implementation. It feels like a lot of implementation details are leaking out of `FileStorage` and into the rest of the app here.

This is especially true if we ever decide we need a different kind of storage than S3. What if, for example, we added an IPFS for Qri FileStorage? (Not that I expect to do that in the near future.) I think retrieving the file’s content (or a stream of the file’s content) probably needs to remain the responsibility of the FileStorage implementation (with the nice-but-not-required performance optimization of redirecting to the actual storage location if it’s publicly accessible, like our S3 buckets).

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +25,11 @@ def show
     }
   end
 
+  def raw
+    @version ||= version_collection.find(params[:id])
+    redirect_to @version.uri, status: 301 and return
```
Mr0grog: I like that turning the `Version#uri` field into a public URL keeps things simple, but given the above issue about what/how different storage implementations might do things, I’m unsure whether that’s really a feasible approach.

- My first gut feeling is that we should be conceiving of `Version#uri` as a URI that we can just pass to a `FileStorage` implementation (maybe it even has a scheme that identifies which implementation?), and different implementations will parse it in different ways to go get what they need. When we serialize a `Version` (i.e. in `Version#as_json`), we should be replacing `uri` with `/api/v0/versions/{uuid}/raw` or `/api/v0/raw/{version_hash}`. Then the route handler at that path figures out what to do by passing the *actual* `uri` from the database to the FileStorage.

- It’s possible we could still store a public URL here, but I feel like it would still need to incorporate FileStore-specific info, since there’s nowhere else to put it. i.e. we could have URLs like `/api/v0/raw/{some-identifier-or-path-or-url-fragment-provided-by-and-passed-back-to-a-filestorage-implementation}`.

    I think the reason I don’t immediately jump to this approach is because it feels like we are freezing part of our API/URL structure into the database, which feels like a coupling that is way too tight. The API is at `v0` for a reason ;)

lib/file_storage/local_file.rb
```diff
@@ -29,13 +32,13 @@ def save_file(path, content, _options = nil)
     end
 
     def url_for_file(path)
-      "file://#{full_path(path)}"
+      polymorphic_url('api_v0_raw', id: path)
```
Mr0grog: See notes above about a) baking URL structure into the DB and b) the way this causes file store implementation work (loading a file from the right location on disk) to be re-implemented outside a file store, and also means parts of the application are tied both to a particular type of storage and a particular location on disk (`LocalFile@directory` is configurable per instance, but not anymore, because it’s hardcoded in your controller).

app/controllers/api/v0/raw_controller.rb
Mr0grog: Also, keep in mind that the storage path is meant to be configured per instance of `LocalFile`, so this is also assuming something (the path being `Rails.root.join('tmp/storage/archive')`) that should probably be determined at configuration time (not compilation time as here). *If nothing else* we should be reading this info from `Archiver.store` instead of hardcoding it here.

jsnshrmn: > So… unless I’m really missing something, this is straight-up not going to work in production or any environment that is not using the LocalFile Filestore implementation.

Nope, not missing anything. My only implementation goal was to serve files that we have, and 4xx errors for those that we don't.

> I think retrieving the file’s content (or a stream of the file’s content) probably needs to remain the responsibility of the FileStorage implementation (with the nice-but-not-required performance optimization of redirecting to the actual storage location if it’s publicly accessible, like our S3 buckets).

I see how I missed something important here on the first swing. It sounds like the goal is to always be able to always serve up the raw body (if it's accessible to our server), even if it means proxying or bridging protocols. We might just choose for it to redirect instead for efficiency. I think I can do that using aws buckets as my non-local reference implementation.

> Also, keep in mind that the storage path is meant to be configured per instance of `LocalFile`, so this is also assuming something (the path being `Rails.root.join('tmp/storage/archive')`) that should probably be determined at configuration time (not compilation time as here). _If nothing else_ we should be reading this info from `Archiver.store` instead of hardcoding it here (but, as noted in the above comment, I’m not sure that’s really good enough either).

Yeah it looks like I dropped a TODO on not hardcoding that path.

Mr0grog: > It sounds like the goal is to always be able to always serve up the raw body (if it's accessible to our server), even if it means proxying or bridging protocols.

Yes! Sorry, I didn’t realize that was not clear. From my perspective, the *problem at hand* is not anything to do with storing files locally — it’s that the stored files might not be publicly accessible (for example, you’d have the same problem with if you used `FileStorage::S3` with a private bucket). Any version record we provide metadata for is one we also *have* to provide the raw body for.

> We might just choose for it to redirect instead for efficiency.

Exactly 👍

Mr0grog:
Also of note here, if you haven’t seen it, is that we already have and use multiple filestorage instances with different configurations. For example we use a different one for storing internal data (like raw import data) than for publicly archived version bodies: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/28b72bcbfe5a4808b2b577a962949a53ac2665bf/app/models/import.rb#L8-L20

jsnshrmn:
I'm finally circling back around to this.  I'm closing this PR since the basic approach is unsuitable.

PR #441, opened 2018-11-30
Quote title text in data migration

Mr0grog:
I failed to properly escape the new `title` field value in the data migration for #436, resulting in https://sentry.io/environmental-data-governance-/db-prod/issues/787187315/activity/

This is pretty simple — just make sure we `Version.connection.quote` the new title.


jsnshrmn:
testing this today.

Mr0grog:
Whoa, we had a lot more malformed titles than I expected! 29,719 versions across 1,369 pages.

PR #452, opened 2018-12-04
Ensure `/pages/{id}/changes` lists correct changes

Mr0grog:
It turns out we were spuriously listing ALL changes (not just the changes for a specified page) at `/api/v0/pages/{page_id}/changes`, which is totally and completely wrong. We also had a wrong test verifying wrong behavior. Don't know how this ever made it in, but it's fixed now.

### DIFF
--- a/app/controllers/api/v0/changes_controller.rb
+++ b/app/controllers/api/v0/changes_controller.rb
@@ -44,7 +44,8 @@ class Api::V0::ChangesController < Api::V0::ApiController
   end
 
   def changes_collection
-    collection = Change.order(created_at: :asc)
+    collection = page ? page.tracked_changes : Change
+    collection = collection.order(created_at: :asc)
     collection = where_in_interval_param(collection, :priority)
     collection = where_in_interval_param(collection, :significance)
     sort_using_params(collection)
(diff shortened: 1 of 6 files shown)


jsnshrmn:
I tried to reproduce the problem in master by importing versions for two pages. Though there are pages and versions, I get zero changes back. I figure I'm skipping some important step. Any idea what it could be? Do I need to do analysis first?
```
{
  "links": {
    "first": "http://192.168.0.96:30000/api/v0/pages/9439d6ef-dc7a-427b-9ad4-85e1d6d46807/changes?chunk=1&chunk_size=100",
    "last": "http://192.168.0.96:30000/api/v0/pages/9439d6ef-dc7a-427b-9ad4-85e1d6d46807/changes?chunk=1&chunk_size=100",
    "prev": null,
    "next": null
  },
  "meta": {
    "total_results": 0
  },
  "data": []
}
```

Mr0grog:
> Do I need to do analysis first?

You need to add annotations (one way of doing that is by running the auto-analysis). Changes are only concretized into actual models when there are annotations to attach to them.

jsnshrmn:
Couldn't get auto analysis working, so I just posted annotations via swagger. Ran into some issues there as well, but I was able to post annotations and reproduce the issue. Now I'm testing the fix.

Issue #458, opened 2018-12-07
Imports should not queue analyses for versions that already exist


Mr0grog:
If an import contains records that are already in our database (especially common if using `?update=merge`), we still trigger auto-analysis for those versions. Since we already have them (and will have presumably already analyzed them, unless they are from a time before we did auto-analysis), it seems like it might not make sense to run analysis on these.

I think we should probably modify things to only queue analysis for versions the import *created* and not for ones it only updated/replaced. (Possible exception: if the `version_hash` or `uri` changed?)

/cc @danielballan @jsnshrmn

PR #459, opened 2018-12-07
Only track a version as added if it was new

Mr0grog:
We were previously tracking versions as added (and thus triggering auto-analysis) if we touched the version in any way. In reality, we should probably only automatically trigger a new analysis if it was actually a new version we *created.*

This is a proposed fix for #458. Should we add some other criteria here? Are there kinds of *updates* that should also trigger analysis? If so, what?

### DIFF
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -100,7 +100,7 @@ class ImportVersionsJob < ApplicationJob
     version.update_different_attribute(save: false)
     version.save
 
-    @added << version
+    @added << version if !existing
   end
 
   def version_for_record(record, existing_version = nil, update_behavior = 'replace')


Mr0grog:
> if we want to do a retrospective analysis of things. e.g. we updated what we're doing in the automated analysis and want to apply those annotations to things from the past

Yep, and I agree that shouldn’t be tied to importing in any way. In fact, we have a Rake task specifically for this :)

```sh
# Args are start & end datetimes, both optional
> rake analyze_changes['2018-12-01','2018-12-10']
```

jsnshrmn:
> we have a Rake task specifically for this :)
Yeah, that would be the place to add options for skipping things with existing annotations or adding annotations to everything within the range. I wasn't able to get this task to execute locally when I was testing annotation-related code recently, so I'll need to figure out what's going on there.

PR #472, opened 2019-01-13
Tag pages with domain, second-level domain

### LINKED ISSUE #470: Automatically tag pages based on domain
After talking organizational tools with the analyst team, we think we should automatically tag pages based on their URL — specifically, we want a tag for the full domain/hostname and a tag for the primary domain. There are two places I can think of for this:

- A callback on the Page model when it is first saved.
- Directly in the import job.

And then the remaining question is what the actual tags should be. I think we should use the same `{type of tag}: {tag name/value}` system we do for Versionista sites here, but what should be the “type” for our two tags?

1. `host` and `domain`, e.g:
    - `host: ntwc.arh.noaa.gov`
    - `domain: noaa.gov`

2. `domain` and `primary-domain`, e.g:
    - `domain: ntwc.arh.noaa.gov`
    - `primary-domain: noaa.gov`

3. Something else???

4. Add fields to `Page` instead of tags. We still need to name the fields, though (see above options).

/cc @danielballan @jsnshrmn

Mr0grog:
When pages are first created, they now automatically tag themselves with two tags:

- `domain:{domain name}` (e.g. `domain:www.nmfs.noaa.gov`)
- `2l-domain:{second-level domain name}` (e.g. `2l-domain:noaa.gov`)

Fixes #470.

---

NOTE: I *could* do a bit more work here and use `around_save` to ensure the tags always match `url` field, even if the `url` changes, but:

- We *shouldn’t* be changing that field
- In a far future where pages have multiple URLs through time, we’d still be *adding* to a list and therefore probably *adding* domain tags

Based on that, complicated code to remove old tags if necessary and add new ones seemed like non-useful complexity at this point. Please say the word if you think I should add support re-writing the domain tags if `url` changes.

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -38,6 +38,7 @@ class Page < ApplicationRecord
   has_many :maintainers, through: :maintainerships
 
   before_create :ensure_url_key
+  after_create :ensure_domain_tags
   before_save :normalize_url
   validate :url_must_have_domain
   validates :status,
@@ -117,6 +118,11 @@ class Page < ApplicationRecord
     update(url_key: Page.create_url_key(url))
   end
 
+  def ensure_domain_tags
+    self.add_tag("domain:#{domain}")
+    self.add_tag("2l-domain:#{second_level_domain}")
+  end
+
   protected
 
   def ensure_url_key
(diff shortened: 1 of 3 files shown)


Mr0grog:
Also please say the word if you think these are bad name choices! It would be nice to avoid rewriting all these tags in the future, but also we need to get this merged this weekend so that we can generate new, actual-domain-based sheets for analysts on Monday.

PR #473, opened 2019-01-14
Remove New Relic gem

### LINKED ISSUE #407: Remove New Relic Gem
**Don’t do this before 2018-10-12 so DNS has time to propagate!**

On Kubernetes (which is now how everything is deployed!), we no longer use any flavor of New Relic, so we should remove the gem.

acaporrini:
Closes #407


Mr0grog:
Thanks, @acaporrini! Looks good :)

Issue #249, opened 2018-03-12
Document how archiving and `ALLOWED_ARCHIVE_HOSTS` work


Mr0grog:
The configuration around data sources is confusing users trying to get everything set up. We should clarify what goes with `ALLOWED_ARCHIVE_HOSTS` and importing/archiving in general.

Mr0grog:
See also #158.

danielballan:
:+1: :+1: :+1: :+1: :+1:

jsnshrmn:
Still relevant and on my to-do list.

Issue #199, opened 2018-01-17
Add content length and content type to Version


Mr0grog:
I recently added `length` (# of bytes) and `content_type` to the source metadata for Versionista imports (edgi-govdata-archiving/web-monitoring-versionista-scraper#63). These both seem like fields that are important enough to be worth promoting to actual fields on the `Version` object. Should we?

Also, are the above names good? Should they change?

- `length`
- `content_type`

/cc @danielballan

danielballan:
I think we should add them and I think the names are good.

Mr0grog:
Removed the question label; I think this is firmly in confirmed/we-want-to-do-it territory now.

Mr0grog:
Thinking about this more, I’d like to add a third field:

- `length` (# of bytes)
- `media_type` (The `type/subtype` portion of the content type/media type — e.g. `text/html`)
- `media_type_parameters` (Additional parameters for the content type — e.g. `foo=bar; something=else`)

I’m feeling like it would be helpful to separate out the parameters from the main type/subtype for easier querying/grouping/etc. I also changed `content_type` to `media_type`, since that has become the official name for the thing in question (as opposed to the name of the HTTP header).

Should we also have a special field for `charset`? Dunno if that’s going overboard and we should just leave it in the `media_type_parameters` field.

Thoughts?

Mr0grog:
Or maybe we should just drop all the parameters altogether; to get them you’d have to pull the raw content.

Mr0grog:
OK, here’s what needs to happen here:

1. Create a migration that adds the above three fields to the `versions` table.

2. Update `ImportVersionsJob` to be able to populate the above fields from similarly named fields (if present) in each import record.

3. Write a [rake task to populate those three fields](https://github.com/edgi-govdata-archiving/web-monitoring-db/tree/master/lib/tasks/data) in existing records (we no longer mix data migrations with Rails schema migrations):
    - For versions whose `uri` starts with `https://edgi-wm-archive.s3.amazonaws.com` or `https://edgi-wm-versionista.s3.amazonaws.com`, you can make a `HEAD` request and trust that the `Content-Length` and `Content-Type` headers are accurate if present.
    - For other versions, you’ll need to do a `GET` request and read the actual content length. I think it’s OK to not try to sniff the content type if there’s no header for it.
[...]

PR #475, opened 2019-01-29
Add `/versions/{uuid}/raw` route

### LINKED ISSUE #431: Add route for raw version bodies

jsnshrmn:
A second attempt to fix #431, based on @Mr0grog's feedback on the questionable solution in #437.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,8 +31,8 @@ MAIL_SENDER='some-email-account@example.com'
 # stores it (see `lib/archiver` for more). That way, we can ensure data is
 # always available to API users from a reliable public location.
 ALLOWED_ARCHIVE_HOSTS='https://edgi-web-monitoring-db.s3.amazonaws.com/ https://edgi-wm-versionista.s3.amazonaws.com/ https://edgi-wm-versionista.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-wm-versionista/ https://edgi-versionista-archive.s3.amazonaws.com/ https://edgi-versionista-archive.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-versionista-archive/'
-
-# Controls URLs that will be served directly rather than proxied
+# OPTIONAL: Additional URLs that will be exposed directly to clients via redirects rather than being proxied by web-monitoring-db.
+#PUBLIC_ARCHIVE_HOSTS='http://static.example.com http://files.example.com'
 PUBLIC_ARCHIVE_HOSTS='https://edgi-web-monitoring-db.s3.amazonaws.com/ https://edgi-wm-versionista.s3.amazonaws.com/ https://edgi-wm-versionista.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-wm-versionista/ https://edgi-versionista-archive.s3.amazonaws.com/ https://edgi-versionista-archive.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-versionista-archive/'
 
 # OPTIONAL: Uncomment & fill in to use S3 for storage instead of your local
--- a/.env.example
+++ b/.env.example
@@ -32,6 +32,9 @@ MAIL_SENDER='some-email-account@example.com'
 # always available to API users from a reliable public location.
 ALLOWED_ARCHIVE_HOSTS='https://edgi-web-monitoring-db.s3.amazonaws.com/ https://edgi-wm-versionista.s3.amazonaws.com/ https://edgi-wm-versionista.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-wm-versionista/ https://edgi-versionista-archive.s3.amazonaws.com/ https://edgi-versionista-archive.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-versionista-archive/'
 
+# Controls URLs that will be served directly rather than proxied
+PUBLIC_ARCHIVE_HOSTS='https://edgi-web-monitoring-db.s3.amazonaws.com/ https://edgi-wm-versionista.s3.amazonaws.com/ https://edgi-wm-versionista.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-wm-versionista/ https://edgi-versionista-archive.s3.amazonaws.com/ https://edgi-versionista-archive.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-versionista-archive/'
+
 # OPTIONAL: Uncomment & fill in to use S3 for storage instead of your local
 # file system
 # AWS_ACCESS_KEY_ID=XXX
--- a/.env.example
+++ b/.env.example
@@ -31,9 +31,6 @@ MAIL_SENDER='some-email-account@example.com'
 # stores it (see `lib/archiver` for more). That way, we can ensure data is
 # always available to API users from a reliable public location.
 ALLOWED_ARCHIVE_HOSTS='https://edgi-web-monitoring-db.s3.amazonaws.com/ https://edgi-wm-versionista.s3.amazonaws.com/ https://edgi-wm-versionista.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-wm-versionista/ https://edgi-versionista-archive.s3.amazonaws.com/ https://edgi-versionista-archive.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-versionista-archive/'
-# OPTIONAL: Additional URLs that will be exposed directly to clients via redirects rather than being proxied by web-monitoring-db.
-#PUBLIC_ARCHIVE_HOSTS='http://static.example.com http://files.example.com'
-PUBLIC_ARCHIVE_HOSTS='https://edgi-web-monitoring-db.s3.amazonaws.com/ https://edgi-wm-versionista.s3.amazonaws.com/ https://edgi-wm-versionista.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-wm-versionista/ https://edgi-versionista-archive.s3.amazonaws.com/ https://edgi-versionista-archive.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-versionista-archive/'
 
 # OPTIONAL: Uncomment & fill in to use S3 for storage instead of your local
 # file system
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -133,7 +133,9 @@ class Api::V0::VersionsController < Api::V0::ApiController
     methods << :change_from_earliest if boolean_param(:include_change_from_earliest)
     options[:methods] = methods
 
-    version.uri = "#{api_v0_version_url(version)}/raw"
-    version.as_json(options)
+    # Don't expose the backend URI, expose the 'raw' route instead.
+    serialized_version = version.as_json(options)
+    serialized_version['uri'] = raw_api_v0_version_url(version)
+    return serialized_version
   end
 end
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -31,10 +31,6 @@ class Api::V0::VersionsController < Api::V0::ApiController
 
     @version ||= version_collection.find(params[:id])
 
-    if ENV['PUBLIC_ARCHIVE_HOSTS']
-      self.public_hosts = ENV['PUBLIC_ARCHIVE_HOSTS']
-    end
-
     def self.public_hosts=(hosts)
       hosts = [] if hosts.nil?
       hosts = hosts.split(' ') if hosts.is_a?(String)
@@ -42,6 +38,10 @@ class Api::V0::VersionsController < Api::V0::ApiController
         raise StandardError, 'Public hosts must be a string or enumerable of strings'
       end
       @public_hosts = hosts
+    end
+
+    if ENV['PUBLIC_ARCHIVE_HOSTS']
+      self.public_hosts = ENV['PUBLIC_ARCHIVE_HOSTS']
     end
 
     def self.public_hosts
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -26,16 +26,19 @@ class Api::V0::VersionsController < Api::V0::ApiController
   end
 
   def raw
-
     @version ||= version_collection.find(params[:id])
+
+    expires_in 1.year, :public => true
 
     if Archiver.public_archive_uri?(@version.uri)
       redirect_to @version.uri, status: 301 and return
     else
-      upstream = Archiver.store.get_file(@version.uri)
+      upstream = Archiver.get_file_from_uri(@version.uri)
+
       mime_type = @version.source_metadata['mime_type']
       mime_type = upstream.content_type if mime_type.nil? || mime_type.empty?
-      send_data(upstream.read, type: mime_type, disposition: 'inline')
+
+      send_data(upstream, type: mime_type, disposition: 'inline')
     end
   end
 
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -30,10 +30,10 @@ class Api::V0::VersionsController < Api::V0::ApiController
 
     expires_in 1.year, :public => true
 
-    if Archiver.public_archive_url?(@version.uri)
+    if Archiver.external_archive_url?(@version.uri)
       redirect_to @version.uri, status: 301 and return
-    else
-      upstream = Archiver.get_file_from_uri(@version.uri)
+    elsif Archiver.store.contains_url?(@version.uri)
+      upstream = Archiver.store.get_file(@version.uri)
 
       # Media type logic mostly cribbed from
       # web-monitoring-db/app/jobs/analyze_change_job.rb
@@ -49,6 +49,8 @@ class Api::V0::VersionsController < Api::V0::ApiController
       end
 
       send_data(upstream, type: media, disposition: 'inline')
+    else
+      raise ActiveRecord::RecordNotFound, "No raw content for #{@version.uuid}."
     end
   end
 
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -35,10 +35,20 @@ class Api::V0::VersionsController < Api::V0::ApiController
     else
       upstream = Archiver.get_file_from_uri(@version.uri)
 
-      mime_type = @version.source_metadata['mime_type']
-      mime_type = upstream.content_type if mime_type.nil? || mime_type.empty?
+      # Media type logic mostly cribbed from
+      # web-monitoring-db/app/jobs/analyze_change_job.rb
+      # Lines 176 to 180 in 658ae8c
+      # TODO: this will eventually be a proper field on `version`:
+      # https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/199
+      meta = @version.source_metadata || {}
+      media = meta['media_type'] || meta['content_type'] || meta['mime_type']
+      if !media && meta['headers'].is_a?(Hash)
+        media = meta['headers']['content-type'] || meta['headers']['Content-Type']
+      elsif !media
+        media = upstream.content_type
+      end
 
-      send_data(upstream, type: mime_type, disposition: 'inline')
+      send_data(upstream, type: media, disposition: 'inline')
     end
   end
 
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -129,6 +129,7 @@ class Api::V0::VersionsController < Api::V0::ApiController
     methods << :change_from_earliest if boolean_param(:include_change_from_earliest)
     options[:methods] = methods
 
+    version.uri = "#{api_v0_version_url(version)}/raw"
     version.as_json(options)
   end
 end
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -28,11 +28,38 @@ class Api::V0::VersionsController < Api::V0::ApiController
   end
 
   def raw
+
     @version ||= version_collection.find(params[:id])
-    upstream = open(@version.uri)
-    mime_type = @version.source_metadata['mime_type']
-    mime_type = upstream.content_type if mime_type.nil? || mime_type.empty?
-    send_data(upstream.read, type: mime_type, disposition: 'inline')
+
+    if ENV['PUBLIC_ARCHIVE_HOSTS']
+      self.public_hosts = ENV['PUBLIC_ARCHIVE_HOSTS']
+    end
+
+    def self.public_hosts=(hosts)
+      hosts = [] if hosts.nil?
+      hosts = hosts.split(' ') if hosts.is_a?(String)
+      unless hosts.is_a?(Enumerable) && hosts.all? {|host| host.is_a?(String)}
+        raise StandardError, 'Public hosts must be a string or enumerable of strings'
+      end
+      @public_hosts = hosts
+    end
+
+    def self.public_hosts
+      @public_hosts || []
+    end
+
+    def self.public_archive_uri?(uri)
+      public_hosts.any? {|base| uri.starts_with?(base)}
+    end
+
+    if public_archive_uri?(@version.uri)
+      redirect_to @version.uri, status: 301 and return
+    else
+      upstream = open(@version.uri)
+      mime_type = @version.source_metadata['mime_type']
+      mime_type = upstream.content_type if mime_type.nil? || mime_type.empty?
+      send_data(upstream.read, type: mime_type, disposition: 'inline')
+    end
   end
 
   def create
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -38,7 +38,7 @@ class Api::V0::VersionsController < Api::V0::ApiController
 
       send_data(upstream, type: media, disposition: 'inline')
     else
-      raise ActiveRecord::RecordNotFound, "No raw content for #{@version.uuid}."
+      raise ApiError::NotFoundError, "No raw content for #{@version.uuid}."
     end
   end
 
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -33,9 +33,8 @@ class Api::V0::VersionsController < Api::V0::ApiController
     if Archiver.external_archive_url?(@version.uri)
       redirect_to @version.uri, status: 301 && return
     elsif Archiver.store.contains_url?(@version.uri)
-      upstream = Archiver.store.get_file(@version.uri)
-      media = version_media_type(@version)
-
+      upstream = Archiver.store.get_file(File.basename(@version.uri))
+      media = version_media_type(@version) || upstream.content_type
       send_data(upstream, type: media, disposition: 'inline')
     else
       raise ApiError::NotFoundError, "No raw content for #{@version.uuid}."
@@ -52,8 +51,6 @@ class Api::V0::VersionsController < Api::V0::ApiController
     media = meta['media_type'] || meta['content_type'] || meta['mime_type']
     if !media && meta['headers'].is_a?(Hash)
       media = meta['headers']['content-type'] || meta['headers']['Content-Type']
-    elsif !media
-      media = upstream.content_type
     end
     media
   end
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -28,10 +28,10 @@ class Api::V0::VersionsController < Api::V0::ApiController
   def raw
     @version ||= version_collection.find(params[:id])
 
-    expires_in 1.year, :public => true
+    expires_in 1.year, public: true
 
     if Archiver.external_archive_url?(@version.uri)
-      redirect_to @version.uri, status: 301 and return
+      redirect_to @version.uri, status: 301 && return
     elsif Archiver.store.contains_url?(@version.uri)
       upstream = Archiver.store.get_file(@version.uri)
 
@@ -149,7 +149,6 @@ class Api::V0::VersionsController < Api::V0::ApiController
     options[:methods] = methods
 
     # Don't expose the backend URI, expose the 'raw' route instead.
-    return version.as_json(options)
-     .update('uri' => raw_api_v0_version_url(version))
+    return version.as_json(options).update('uri' => raw_api_v0_version_url(version))
   end
 end
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -147,8 +147,7 @@ class Api::V0::VersionsController < Api::V0::ApiController
     options[:methods] = methods
 
     # Don't expose the backend URI, expose the 'raw' route instead.
-    serialized_version = version.as_json(options)
-    serialized_version['uri'] = raw_api_v0_version_url(version)
-    return serialized_version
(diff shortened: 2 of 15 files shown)


jsnshrmn:
So I feel like this is a reasonable redirect-less solution that I was able to get to pretty quickly. I've noodled around locally with redirects, and I end up touching a bunch of stuff and end up sticking a mix of file storage stuff into records upon creation. It ends up looking messy/wrong to me.

In #431 it sounds like redirects are maybe a "very nice to have" instead of a "must have." Maybe I just stay out of file storage for now and we proxy all raw versions? How does this look so far to you, @Mr0grog?

jsnshrmn:
My thinking on storage and redirects:

We could have versions potentially stored with different routeability/accessibility. e.g. we might add a new non-routeable storage type and begin importing new versions to it (where we couldn't do redirects) while keeping existing versions on accessible buckets (where we could do redirects). To know whether a given version needed to be proxied or could be redirected, I'm playing with sticking something (latest thought is a 'uri_acl' string) in the record upon import. Am I thinking about this correctly/making the right assumptions?

jsnshrmn:
I suppose it could also be done in configuration. We could have a PUBLIC_ARCHIVE_HOSTS environment variable that had a redirect-able white list in it.

jsnshrmn:
Did a quick implementation of configuration via PUBLIC_ARCHIVE_HOSTS.

jsnshrmn:
I need to add tests here!

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +27,41 @@ def show
...
+      self.public_hosts = ENV['PUBLIC_ARCHIVE_HOSTS']
+    end
+
+    def self.public_hosts
+      @public_hosts || []
+    end
+
+    def self.public_archive_uri?(uri)
+      public_hosts.any? {|base| uri.starts_with?(base)}
+    end
```
Mr0grog: Are all these inline methods meant to ultimately be here or is this just proof-of-concept? This definitely isn’t mergeable like this.

These methods should at least be part of the class definition, if not separated into a concern/mixin or even a separate service class or module (like `Archiver`, perhaps—or maybe this logic belongs *in* archiver).

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +27,41 @@ def show
...
+      hosts = [] if hosts.nil?
+      hosts = hosts.split(' ') if hosts.is_a?(String)
+      unless hosts.is_a?(Enumerable) && hosts.all? {|host| host.is_a?(String)}
+        raise StandardError, 'Public hosts must be a string or enumerable of strings'
+      end
+      @public_hosts = hosts
+    end
+
+    if ENV['PUBLIC_ARCHIVE_HOSTS']
+      self.public_hosts = ENV['PUBLIC_ARCHIVE_HOSTS']
```
Mr0grog: It *seems* to me like this concept of “public archive hosts” could be a superset of what’s in `ALLOWED_ARCHIVE_HOSTS`. If the `uri` points to a place that we didn’t archive locally because it’s in an accessible, public location, that might be a good indicator that we should redirect to it. That said, I can see how those are technically separate concepts, and there could be some future situation where this wouldn’t be a strict superset of that. I guess I’m not really sure if that’s actually a useful distinction in practice.

Also, I *generally* prefer keeping things like environment variables confined to the `config` directory (usually by setting up things in an initializer), that way it’s slightly more possible to keep track of what env vars are used where.

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +27,41 @@ def show
...
+    end
+
+    def self.public_archive_uri?(uri)
+      public_hosts.any? {|base| uri.starts_with?(base)}
+    end
+
+    if public_archive_uri?(@version.uri)
+      redirect_to @version.uri, status: 301 and return
+    else
+      upstream = open(@version.uri)
```
Mr0grog: See above about feature and security issues with `open-uri`.

If the `uri` is not public, I think the right thing here might actually be calling `Archiver.store.get_file()`.

app/controllers/api/v0/versions_controller.rb
```diff
@@ -118,6 +155,7 @@ def serialize_version(version, options = {})
     methods << :change_from_earliest if boolean_param(:include_change_from_earliest)
     options[:methods] = methods
 
+    version.uri = "#{api_v0_version_url(version)}/raw"
```
Mr0grog: Should be `raw_api_v0_version_url(version)`, no string substitution or manual path writing necessary.

I would also feel a *lot* better if we didn’t mutate and dirty the actual model object here (otherwise, once you’ve called this method, you’d better be careful not to do anything that might cause it to save and you also can’t read the `uri` field anymore). Instead, we should modify the hash returned from `as_json`.

config/routes.rb
```diff
@@ -28,6 +28,10 @@
       end
 
       resources :versions, only: [:index, :show], format: :json
+      resources :versions, only: [:show] do
+        get 'raw', on: :member
+      end
```
Mr0grog: Why isn’t this part of the above `resources :versions` call?

It might also make sense to specify `format: false` here, since it doesn’t really make sense to ask for a particular content type on the `raw` endpoint:

```rb
get 'raw', on: member, format: false
```

That prevents a route like `/api/v0/versions/abc123/raw.json` from working — you’d have to remove the `.json` part.

config/routes.rb
```diff
@@ -28,6 +28,10 @@
...
+      resources :raw, only: [:show]
```
Mr0grog: This is setting up a route to the `show` method on the controller at `app/controllers/api/v0/raw`, which doesn’t exist in this PR. Is that something you were planning to add?

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +27,41 @@ def show
...
+      public_hosts.any? {|base| uri.starts_with?(base)}
+    end
+
+    if public_archive_uri?(@version.uri)
+      redirect_to @version.uri, status: 301 and return
+    else
+      upstream = open(@version.uri)
+      mime_type = @version.source_metadata['mime_type']
+      mime_type = upstream.content_type if mime_type.nil? || mime_type.empty?
+      send_data(upstream.read, type: mime_type, disposition: 'inline')
```
Mr0grog: Should probably set some long cache headers here, too.

app/controllers/api/v0/versions_controller.rb
```diff
@@ -1,3 +1,5 @@
+require 'open-uri'
```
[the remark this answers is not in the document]
jsnshrmn: This comment led me to experiment with snapshot URIs like `/etc/ssh/ssh_host_rsa_key` I will definitely switch away from `open-uri`.

app/controllers/api/v0/versions_controller.rb
jsnshrmn: Yeah, this is definitely a sketch that needs to be cleaned up.

app/controllers/api/v0/versions_controller.rb
jsnshrmn: Ah, thank you for this. This felt like an icky choice when I did it, but I wasn't sure what the correct choice was.

config/routes.rb
jsnshrmn: This was a leftover from copying the few usable bits over from my first pr on this issue. to start this one. Will cleanup.

config/routes.rb
jsnshrmn: same as above. will cleanup.

app/controllers/api/v0/versions_controller.rb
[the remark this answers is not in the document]
Mr0grog: Also: the idea I outlined around not storing storage-config, but just storage names, in the DB also helps secure things better—we can ensure that any address in a file store is always scoped to a given directory.

app/controllers/api/v0/versions_controller.rb
Mr0grog: `rake routes` is a very useful tool if you aren’t totally up on how Rails names these things :)

app/controllers/api/v0/versions_controller.rb
jsnshrmn: Done. Redirects are now cacheable as well. Is that desirable?

app/controllers/api/v0/versions_controller.rb
jsnshrmn: That seems right to me too. Done.

app/controllers/api/v0/versions_controller.rb
jsnshrmn: Makes sense. I setup the public archive check method to test for public or allowed hosts and made the environment variable optional. I also moved the environment variable handling into the archiver initializer since I moved the related checking/fetching methods into archiver as well.

app/controllers/api/v0/versions_controller.rb
jsnshrmn: I moved pretty much everything into Archiver.

app/controllers/api/v0/versions_controller.rb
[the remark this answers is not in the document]
jsnshrmn: Moved  those raw proxy requests over to HTTParty as advised, and reduced the circumstances under which requests will get made.

app/controllers/api/v0/versions_controller.rb
jsnshrmn: Yeah, I used `rake routes` while setting up my initial PR for this issue, but it seemingly fell out of my brain by the time I came around to this new approach. 😳This should be in a happier place now.

jsnshrmn:
This still needs tests, but it would be nice to get another look to see if the approach is going to be satisfactory.

.env.example
```diff
@@ -23,8 +23,10 @@ MAIL_SENDER='some-email-account@example.com'
 # MAIL_SMTP_PASSWORD='XXX'
 # MAIL_SMTP_TLS='true'
 
-# Controls URLs that won't be downloaded and re-hosted when importing versions
+# Controls URLs that won't be downloaded and re-hosted when importing versions. These will be exposed directly to clients via redirects rather than being proxied by web-monitoring-db.
 ALLOWED_ARCHIVE_HOSTS='https://edgi-web-monitoring-db.s3.amazonaws.com/ https://edgi-wm-versionista.s3.amazonaws.com/ https://edgi-wm-versionista.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-wm-versionista/ https://edgi-versionista-archive.s3.amazonaws.com/ https://edgi-versionista-archive.s3-us-west-2.amazonaws.com/ https://s3-us-west-2.amazonaws.com/edgi-versionista-archive/'
+# OPTIONAL: Additional URLs that will be exposed directly to clients via redirects rather than being proxied by web-monitoring-db.
+#PUBLIC_ARCHIVE_HOSTS='http://static.example.com http://files.example.com'
```
Mr0grog: I know I suggested this direction, but the more I think about it, the more complex it feels. We already reliably confuse people with `ALLOWED_ARCHIVE_HOSTS` (see #249) and this just adds to the complexity there and adds a lot to `Archiver`. I’m wondering if a better model might be to indicate whether the storage is public. So `FileStorage::LocalFile` would always be private, but `FileStorage::S3` might check `@acl`.

Thoughts? If you disagree, go ahead and stick with this. If you do, though, I think it needs more explanation to clarify that it’s essentially *additive* on top of `ALLOWED_ARCHIVE_HOSTS`.

lib/archiver/archiver.rb
```diff
@@ -29,6 +29,19 @@ def self.allowed_hosts
     @allowed_hosts || []
   end
 
+  def self.public_hosts=(hosts)
+    hosts = [] if hosts.nil?
+    hosts = hosts.split(' ') if hosts.is_a?(String)
+    unless hosts.is_a?(Enumerable) && hosts.all? {|host| host.is_a?(String)}
+      raise StandardError, 'Public hosts must be a string or enumerable of strings'
+    end
```
Mr0grog: If you keep this (see comment on the `.env.example` file), I think this list parsing logic should be extracted so we don’t duplicate what’s already in the `allowed_hosts` setter.

lib/archiver/archiver.rb
```diff
@@ -81,6 +98,20 @@ def self.external_archive_url?(url)
     allowed_hosts.any? {|base| url.starts_with?(base)}
   end
 
+  def self.get_file_from_uri(uri)
+    # HTTP GET the file if its URI belongs to a public host.
+    if public_archive_url?(uri)
```
Mr0grog: I don’t think we ever execute this branch of logic in the current implementation. Is there an important use case you’re anticipating here? Otherwise, let’s drop it.

app/controllers/api/v0/versions_controller.rb
Mr0grog: I don’t think there should be a problem with that. That said, I’m not even sure if most clients respect cache headers on redirects (or how they cross them with what different 3xx codes mean if they do respect them). Are you concerned that we might decide in the future to move everything sometime and this would cause us trouble?

jsnshrmn: I mostly just checking in, because I noticed that firefox will recheck the original URL on a hard reload if the redirect wasn't cacheable, but will just go straight to the 301 target if the redirect was cacheable. I imagine that the behavior with the current response varies quite a bit in different user agents. If we do end up having an unforeseen move of assets from the current s3 buckets, we can always leave the current bucket up with redirects.

lib/archiver/archiver.rb
jsnshrmn: I was just making sure this method works as advertised in case there's some other use for it in the future. No idea what that use might be, but if this is supposed to be a public API, I'd leave it in for completeness; if not I'll drop it.

.env.example
jsnshrmn: Yeah, we definitely have plenty of complexity that leaks into configuration. Let me think on this.

app/controllers/api/v0/versions_controller.rb
Mr0grog: > Firefox will recheck the original URL on a hard reload if the redirect wasn't cacheable, but will just go straight to the 301 target if the redirect was cacheable.

That’s fascinating and good to know!

> I mostly just checking in… If we do end up having an unforeseen move of assets from the current s3 buckets, we can always leave the current bucket up with redirects.

Cool, so I think I think we’re good here. 👍

lib/archiver/archiver.rb
Mr0grog: I think I might rename this method to something more like `get_archived_file()`, which wouldn’t imply this would ever go out to the internet, and you might not worry so much about needing to have something like this for completeness’s sake. That would also make it a clearer analog to `Archiver.archive()`.

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +25,23 @@ def show
...
 
+  def raw
+    @version ||= version_collection.find(params[:id])
+
+    expires_in 1.year, :public => true
+
+    if Archiver.public_archive_url?(@version.uri)
+      redirect_to @version.uri, status: 301 and return
+    else
+      upstream = Archiver.get_file_from_uri(@version.uri)
```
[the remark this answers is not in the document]
jsnshrmn: It looks like there's some magic happening here. Actually implementing version_hash as the ETag value is looking more complicated than I first thought it would be. Unless I'm missing something, it looks like we'd either need to implement our own cache_key method for versions, or else manually `response.set_header` and do our own conditional response handling. It does look like there are already conditional responses based on automagical weak Etags.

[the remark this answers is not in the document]
Mr0grog: > It does look like there are already conditional responses based on automagical weak Etags

I guess that doesn't seem surprising. But it also seems like the right etag (and actually, maybe it should be strong?) in this case is our `version_hash`, so we’d rather have that than what Rails automatically gins up for us.

All that said, it looks like it’s ultimately setting the etag to an MD5 hex digest of whatever you pass in, which is a little pointless in our case, but probably not harmful. As long as it’s deterministic, all good?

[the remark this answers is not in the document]
Mr0grog: (I should say, MD5 hex digest of the explicit etag we pass in + last modified date + anything extra the controller specified should be in all etags.)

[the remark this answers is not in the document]
Mr0grog: Way more special and less direct than I’d realized whenever I’d used `stale?` before!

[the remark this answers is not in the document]
jsnshrmn: > As long as it’s deterministic, all good?

Works for me!

jsnshrmn:
This round of feedback caused me to pare things down as much as possible, to the point where `PUBLIC_ARCHIVE_HOSTS` has disappeared. I *think* this is about as straightforward as my "don't fiddle with the storage layer" approach can get. Obvious outstanding things are tests and updated docs for  `ALLOWED_ARCHIVE_HOSTS` (maybe closing #249 while I'm here)

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +25,35 @@ def show
...
+      media = meta['media_type'] || meta['content_type'] || meta['mime_type']
+      if !media && meta['headers'].is_a?(Hash)
+        media = meta['headers']['content-type'] || meta['headers']['Content-Type']
+      elsif !media
+        media = upstream.content_type
+      end
+
+      send_data(upstream, type: media, disposition: 'inline')
+    else
+      raise ActiveRecord::RecordNotFound, "No raw content for #{@version.uuid}."
```
Mr0grog: I think it would be good not to co-opt an ActiveRecord error just to get a 404 here. If you want to raise, the `ApiError` module is probably what you should use. (We co-opt this error in `Change#find_by_api_id` only because that is doing an actual ActiveRecord lookup, and is meant to work and feel as if it is an ActiveRecord finder method from the outside.)

Mr0grog:
Any updates on this, @jsnshrmn?

jsnshrmn:
Not yet, but I will keep you posted!

jsnshrmn:
Not yet, but I will keep you posted!

Mr0grog:
@jsnshrmn Do you think you’re going to have time to clean this up this week? If not, I think I may try and work on it.

jsnshrmn:
It seems unlikely to happen this week. I've been swamped since April 15, and haven't gotten to it yet.

Mr0grog:
@jsnshrmn are you actively working on this?

jsnshrmn:
Yep, I initially thought I would work on it in between contracts, but I've been continuously booked up. I decided to come back to it as a Saturday project instead of leaving it sitting any longer. Sorry for leaving it to rot on the vine for so long.

jsnshrmn:
Setup a fresh working dev environment for this. I hadn't done that previously since I was just resolving merge conflicts and looking at the linter output.

Mr0grog:
Hey @jsnshrmn, I just wanted to check in again and see how this is going. Are you still actively working, is there anything you’re stuck on?

I also want to make sure we don’t start getting a lot of feature or complexity creep in here again. I saw you added libmagic support, and it looks like you’re using it to sniff content types. I think that might not really be needed — we should just make sure the metadata in the version record is good. I think it’s OK to leave sniffing up to the client if we don’t actually know the content type.

(We also still have #199 to fix, and when that’s done, filling in the `content_type` field in instances where other info is missing can probably be a one time task. We should also be able to do that by checking the `Content-Type` header in S3, which we have *always* been careful to set, even before we started putting it in `source_metadata`).

jsnshrmn:
hi there. yeah, still working on this intermittently. I added the packet sniffing to get a content type for local storage, as I have to set something to `send_data` otherwise it defaults to application/octet-stream. grabbing the `content_type` from s3 has worked very well when using that storage type. looking at #199 I can see that sniffing filetypes here would be a totally redundant addition of a dependency. I'll strip the libmagic stuff out out and just put in a fixed value in as a placeholder for now while I work out the other bits.

.env.example
jsnshrmn: `PUBLIC_ARCHIVE_HOSTS` is no more.

app/controllers/api/v0/versions_controller.rb
jsnshrmn: this has been dispensed with

lib/archiver/archiver.rb
jsnshrmn: dispensed with this

lib/archiver/archiver.rb
jsnshrmn: dispensed with this.

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +25,35 @@ def show
...
 
+  def raw
+    @version ||= version_collection.find(params[:id])
+
+    expires_in 1.year, :public => true
+
+    if Archiver.external_archive_url?(@version.uri)
+      redirect_to @version.uri, status: 301 and return
+    elsif Archiver.store.contains_url?(@version.uri)
+      upstream = Archiver.store.get_file(@version.uri)
```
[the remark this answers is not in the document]
jsnshrmn: I'm now doing `File.basename(@version.uri)`, which works happily on local storage, but I haven't done any consideration / checking on s3.

jsnshrmn:
Now that I've cut out the problematic bits, there's not much left. This seems to be working in some basic spot checks with local storage. I haven't considered s3 storage, created a registry of storage objects in archiver, or written any tests yet.

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +25,42 @@ def show
...
+    expires_in 1.year, public: true
+
+    if Archiver.external_archive_url?(@version.uri)
+      redirect_to @version.uri, status: 301 && return
+    elsif Archiver.store.contains_url?(@version.uri)
+      # Get the file
+      filename = File.basename(@version.uri)
+      upstream = Archiver.store.get_file(filename)
+      # Try to get the filetype, fall back on binary.
+      type = version_media_type(@version) || upstream.content_type || 'application/octet-stream'
```
Mr0grog: `upstream.content_type` is no longer a thing, right?

jsnshrmn: If I recall correctly, it is in the s3 file objects. I'll verify that when I go check the s3 functionality. I named the (now removed) singleton for local files to align with method of s3 objects. I can drop this anyway if you want, since it sounds like the plan is to backfill versions with info.

Mr0grog: `get_file` should return a string, not annotated with anything special: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/459c23211455966826b1470562e1ebe81294d431/lib/file_storage/s3.rb#L25-L27

(Maybe it would be a good improvement to return some more special kind of object, but that should probably be out of scope here.)

Mr0grog: Or: none of the methods in `FileStorage::S3` should give you back any S3 types. It should have the exact same public API as `FileStorage::LocalFile`.

jsnshrmn: Interesting. I may have be recalling some bizarre behavior that I myself introduced on a previous iteration of this. I'll just kill `content_type` as requested.

jsnshrmn:
I *think* this is in a happier place now.

app/controllers/api/v0/versions_controller.rb
```diff
@@ -119,6 +156,7 @@ def serialize_version(version, options = {})
     methods << :change_from_earliest if boolean_param(:include_change_from_earliest)
     options[:methods] = methods
 
-    version.as_json(options)
+    # Don't expose the backend URI, expose the 'raw' route instead.
+    version.as_json(options).update('uri' => raw_api_v0_version_url(version))
```
Mr0grog: Wondering if we should 

```suggestion
    result = version.as_json(options)
    result.update('uri' => raw_api_v0_version_url(version)) unless version.uri && Archiver.external_archive_url?(version.uri)
    result
```

Probably not a huge deal either way, but would *slightly* reduce the amount of traffic we send ourselves.

(See other comment for more on the `version.uri &&` part.)

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +25,43 @@ def show
     }
   end
 
+  def raw
+    @version ||= version_collection.find(params[:id])
+
+    expires_in 1.year, public: true
+
+    if Archiver.external_archive_url?(@version.uri)
```
Mr0grog: JUST realized we have some old data with no `version.uri`. (Only ever happened with Versionista: in some rare cases it would store nothing, and in a very tiny number of cases we lost some old stored responses before we started archiving them ourselves, since it tops out at storing the 50 latest per page.)

Anyway. We should make sure `Archiver.external_archive_url?` and `FileStorage::<Whatever>.contains_url?` handle `nil` inputs and/or bail out early with a 404 if `@version.uri.nil?`.

Mr0grog: Alternatively, I guess we could allow an empty string for `version.uri` and migrate the DB to default this column to `""` and disallow NULLs. But having NULLs in the DB seems clearer to me. Dunno ¯\\\_(ツ)\_/¯

jsnshrmn: Added a check for null URIs to the raw view. Definitely prefer to leave nulls in the database when they are a semantic reality of the data,

app/controllers/api/v0/versions_controller.rb
```diff
@@ -25,6 +25,43 @@ def show
...
+      redirect_to @version.uri, status: 301
+    elsif Archiver.store.contains_url?(@version.uri)
+      # Get the file
+      path = URI.parse(@version.uri).path
+      filename = File.basename(path)
+      upstream = Archiver.store.get_file(filename)
```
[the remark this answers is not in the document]
jsnshrmn: I've taken a swing at adding url handling to get_file.  I updated both local and s3 file handling. Not sure if I handled it the right way, but I see what you're saying about leveraging the existing logic inside of file storage to figure out what to do with urls. Maybe it would be better to split that stuff out into a `get_file_from_url` method instead of handling two different types of arguments in `get_file`? Very open to feedback on that.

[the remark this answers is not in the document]
jsnshrmn: the more I think about my 1st cut on this the more I don't like it, since I've added complexity to `get_file` where the raw route is the only thing using it. Also, tests that are passing locally are failing in circle ci. Maybe I'll go ahead and split that out on the next pass.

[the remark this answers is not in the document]
Mr0grog: Ahhhh, that does seem complex. FWIW, I was thinking something like:

```rb
def get_file(path)
  # If it's a URL, get the path
  path = path[7..-1] if path.starts_with? 'file://'

  # If it's absolute, make sure it's in this storage's directory
  absolute = if path.starts_with?('/')
    raise ArgumentError, "The path '#{path}' does not belong to this storage object" unless path.starts_with?(File.join(directory, ''))
    path
  else
    full_path(path)
  end

  File.read(full_path(path))
end

…but not sure if I’m missing some major issue here.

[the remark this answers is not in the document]
Comment (Mr0grog) [reply app/controllers/api/v0/versions_controller.rb]: Update: fixed a typo above!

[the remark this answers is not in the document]
Comment (Mr0grog) [reply app/controllers/api/v0/versions_controller.rb]: Ok, let’s not delay this for something that seems like it might be introducing confusion or complexity. Why don’t you go ahead and pull the `get_file` changes back out and we can probably call this PR ready to go?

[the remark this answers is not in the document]
Comment (jsnshrmn) [reply app/controllers/api/v0/versions_controller.rb]: Went ahead and reverted that commit.

File: test/controllers/api/v0/versions_controller_test.rb [hunk 19]
```diff
@@ -192,6 +202,42 @@ class Api::V0::VersionsControllerTest < ActionDispatch::IntegrationTest
...
+  end
+
+  test 'can return raw response body for a single version' do
+    Archiver.allowed_hosts = []
+    Archiver.store = FileStorage::S3.new(
+      key: 'whatever',
+      secret: 'test',
+      bucket: 'test-bucket',
+      region: 'us-west-2'
+    )
```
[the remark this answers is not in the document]
jsnshrmn: I tried to add a test here, but I got stuck on the appropriate way to set the uri:
fixtures doesn't allow variables in the source yaml, and the file URIs are absolute paths, meaning that I kind of need to know the rails root, but can't use it in the fixture to set the uri.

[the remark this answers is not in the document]
Mr0grog: > fixtures doesn't allow variables in the source yaml, and the file URIs are absolute paths

The fixtures are actually ERB templates, so you can use `<%= expression.here %>`, e.g: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/55ffb05bd63074f780253ed515686e1c8cd9619b/test/fixtures/invitations.yml#L7-L10

You also don’t necessarily need to use a fixture! I think I overused them in the early tests I wrote, and that was bad.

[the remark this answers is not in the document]
jsnshrmn: > The fixtures are actually ERB templates

Didn't know that! I went ahead and stuck with fixtures, since that is consistent with the other tests.

jsnshrmn:
So, I'm seeing test failures for the import versions job in circleci that I'm not sure what to do with:
https://circleci.com/gh/edgi-govdata-archiving/web-monitoring-db/1428
it seems to be rejecting the file uri, but I'm not sure where the requirement for uri type is being set.
There is another issue that I thought was happening because I was futzing with an existing version used in other tests, so I created new versions for my tests. The error is still there, though.

Mr0grog:
Well, for starters, you’ve got a real in problem in that all your new fixtures belong to the same page and have the same `source_type` + `capture_time`. That’s a validation error, and the system normally goes through a lot of effort to make that case impossible, but you’re skipping all that validation when you inject a fixture.

The error `'<url>' Must be HTTP, HTTPS or Generic` is coming from HTTParty, and is likely originating here where we a try and archive a URL on import: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/55ffb05bd63074f780253ed515686e1c8cd9619b/app/jobs/import_versions_job.rb#L87-L89

We’d be getting there because which version Postgres returns and we try and merge with is somewhat random at this point (because you have four versions all matching the same query conditions that are supposed to be unique, see above). If it picks the one you’ve made with a `file:` URI, it’s going to try and re-archive because `!Archiver.already_archived?(version.uri)` will always be true: you’ve set up the `file:` URL to be at a different path than the Archiver’s storage in test.

Your fixture:

https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/43c69f7f06d8192cf97aad1c1f6b60d4d412d0aa/test/fixtures/versions.yml#L168

The location being used in the import job test (no custom storage for that test):

https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/43c69f7f06d8192cf97aad1c1f6b60d4d412d0aa/config/initializers/archiver.rb#L21-L24

Does that make sense?

test/fixtures/versions.yml
```diff
@@ -139,3 +139,37 @@ page1_v5:
     site_id: 1
     page_id: 11
     version_id: 115
+
+page3_v1:
+  page: home_page
```
Mr0grog: Also, please use a different actual `page` from the one we are using for `page1_v*` fixtures when making a fixture named `page3_v*`. Same for the other new fixtures here.

Mr0grog:
Hey @jsnshrmn, just checking in. Are you still actively working on this?

jsnshrmn:
I stepped away from this for a little while, but  I'm planning to come back to it.

Mr0grog:
OK. FYI, I’m probably going to merge #628 tomorrow, and that’s a pretty major update. I don’t think anything you’re doing here should be an issue in Rails 6, but worth noting in case you do have issues updating.

PR #487, opened 2019-02-14
Only auto-analyze versions that were different

Mr0grog:
If a version had the same content as the version that preceded it, then there's not a whole lot of point in auto-analyzing it -- there's nothing useful in the change between it and preceeding version or between it and the first version (because that change will be the same as between the preceeding version and the first version). Don't queue these versions for auto-analysis.

Noticed this while dealing with the massive number of analyses we queued up when I backfilled a month’s worth of Wayback versions yesterday. It wouldn’t have made a *huge* difference (maybe 10%; I didn’t do a deep analysis), but still seems like a worthwhile optimization.


Mr0grog:
Going ahead and merging this in, since no feedback.

PR #490, opened 2019-02-22
Upgrade Ruby from v2.5.1 to v2.6.3

Mr0grog:
Ruby 2.6 has a bunch of speed improvements, plus some nice new syntactical bits. It also has a new endless ranges feature, but I'm not sure yet whether that could replace our `NumericInterval` class.


Mr0grog:
The new endless ranges don’t really have the same open-ended/closed-ended features `NumericInterval` does, so I don’t think it quite works as a replacement (and we’d have to add our [string notation parsing](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/4e8d9b1b8dc2f22a75b959ea1503739afb5d569a/app/lib/numeric_interval.rb#L7-L22) back in top anyway).

Mr0grog:
The latest Ruby is also now 2.6.3 (this has been sitting for way too long).

Mr0grog:
Merging; this just needs to go in at this point without waiting for other pull requests.

PR #491, opened 2019-02-22
Upgrade bundler to v2

Mr0grog:
Manage gems with v2 of Bundler. This is a bit of a major upgrade, and since bundler is what manages our dependencies, people will have to update their local Bundler install manually. That should be as simple as `gem install bundler` on the command line.

NOTE: this depends on #490.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,11 @@ jobs:
           key: web-monitoring-db-{{ arch }}-{{ checksum "Gemfile.lock" }}
 
       # Bundle install dependencies
-      - run: bundle install --path vendor/bundle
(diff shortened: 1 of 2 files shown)


Mr0grog:
I rebased on the latest master since this has gotten a bit stale.

Mr0grog:
This wasn’t previously supported by Heroku, but now is: https://devcenter.heroku.com/articles/ruby-support#libraries

Need to make sure everything builds right in Docker.

Mr0grog:
Rebased on master and tested Docker builds (which are fine). This is ready to go, thought I’d like to merge #536 first for the sake of a nice organized history :)

PR #530, opened 2019-05-16
Add system tests for user lifecycle

bensheldon:
- Sets up Rails system tests, including capybara and capybara-email
- Creates a first system test that tests inviting a new user, confirmation, promotion to admin, demotion, and deletion
- ~~Within the development environment, replaces Sendmail delivery with letter_opener gem~~ unrelated to the system testing and removed
- Creates a binstub to start the job worker process during development

Note: I am using `rack_test` instead of Selenium Chromedriver because that has more complicated setup than I wanted to tackle right now.

This is a first step towards #34 by wrapping the existing auth system with tests.

### DIFF
--- a/bin/worker_dev
+++ b/bin/worker_dev
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+
+QUEUE=* VERBOSE=1 RESQUE_PRE_SHUTDOWN_TIMEOUT=10 RESQUE_TERM_TIMEOUT=10 bundle exec rake environment resque:work
--- a/test/system/user_test.rb
+++ b/test/system/user_test.rb
@@ -1,0 +1,103 @@
+require "application_system_test_case"
+
+class UsersTest < ApplicationSystemTestCase
+  test "Invite, register, confirm, promote, demote, delete" do
+    admin = users(:admin_user)
+    viewer_email = 'viewer@example.com'
+
+    Capybara.using_session(:admin) do
+      visit root_path
+
+      click_on "Login"
+
+      fill_in 'Email', with: admin.email
+      fill_in 'Password', with: 'testpassword'
+      click_on 'Log in'
+
+      assert page.has_content? "Logged in as #{admin.email}"
+      click_on 'Admin'
+
+      perform_enqueued_jobs do
+        fill_in 'E-mail:', with: viewer_email
+        click_on 'Generate Invitation'
+      end
+    end
+
+    Capybara.using_session(:viewer) do
+      open_email(viewer_email)
+      current_email.find('a[href*="invitation"]').click
+      clear_emails
+
+      fill_in 'Password', with: 'testpassword'
+      fill_in 'Password confirmation', with: 'testpassword'
+      click_on 'Sign up'
+
+      open_email(viewer_email)
+      current_email.click_link 'Confirm my account'
+      clear_emails
+
+      fill_in 'Email', with: viewer_email
+      fill_in 'Password', with: 'testpassword'
+      click_on 'Log in'
+
+      assert page.has_content? "Logged in as viewer@example.com"
+    end
+
+    Capybara.using_session(:admin) do
+      visit admin_path
+      assert page.has_content? "Logged in as #{admin.email}"
+
+      user_row = page.all('tr').find { |tr| tr.has_content? viewer_email }
+      within user_row do
+        click_on 'Promote to admin'
+      end
+    end
+
+    Capybara.using_session(:viewer) do
+      visit root_path
+
+      # TODO: Signing in again should not be necessary.
+      click_on 'Login'
+      fill_in 'Email', with: viewer_email
+      fill_in 'Password', with: 'testpassword'
+      click_on 'Log in'
+
+      assert page.has_content? 'Admin'
+    end
+
+    Capybara.using_session(:admin) do
+      visit admin_path
+      assert page.has_content? "Logged in as #{admin.email}"
+
+      user_row = page.all('tr').find { |tr| tr.has_content? viewer_email }
+      within user_row do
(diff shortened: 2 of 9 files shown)


bin/worker_dev
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+
+QUEUE=* VERBOSE=1 RESQUE_PRE_SHUTDOWN_TIMEOUT=10 RESQUE_TERM_TIMEOUT=10 bundle exec rake environment resque:work
```
Mr0grog: Nice, this is super useful. I always wind up consulting `Procfile` to remember what to run. 👍

test/system/user_test.rb
```diff
@@ -0,0 +1,103 @@
...
+    Capybara.using_session(:admin) do
+      visit root_path
+
+      click_on "Login"
+
+      fill_in 'Email', with: admin.email
+      fill_in 'Password', with: 'testpassword'
+      click_on 'Log in'
+
+      assert page.has_content? "Logged in as #{admin.email}"
```
Mr0grog: Would you please add a message to these assertions, so it’s a little clearer what’s happening if they fail? (Same with the other ones below.)

test/system/user_test.rb
```diff
@@ -0,0 +1,103 @@
+require "application_system_test_case"
+
+class UsersTest < ApplicationSystemTestCase
```
Mr0grog: Please make sure the file and class names match (the class is plural).

test/system/user_test.rb
```diff
@@ -0,0 +1,103 @@
...
+      user_row = page.all('tr').find { |tr| tr.has_content? viewer_email }
+      within user_row do
+        click_on 'Promote to admin'
+      end
+    end
+
+    Capybara.using_session(:viewer) do
+      visit root_path
+
+      # TODO: Signing in again should not be necessary.
```
Mr0grog: Oh, this is super weird! But I definitely cannot figure out what’s going wrong here either. :(

test/system/user_test.rb
```diff
@@ -0,0 +1,103 @@
+require "application_system_test_case"
+
+class UsersTest < ApplicationSystemTestCase
+  test "Invite, register, confirm, promote, demote, delete" do
+    admin = users(:admin_user)
+    viewer_email = 'viewer@example.com'
+
+    Capybara.using_session(:admin) do
```
Mr0grog: Would you mind adding a comment at the start of each of these blocks describing what it’s doing conceptually? This is a long function that flips between sessions, and could use description to make it easier to follow.

bin/worker_dev
bensheldon: 🎆

test/system/user_test.rb
bensheldon: I've added assertion messages to all `assert` and `refute` calls.

test/system/user_test.rb
bensheldon: Whoops! Renamed.

test/system/user_test.rb
bensheldon: Fixed it. The problem is that when clicking through the URL in the mailer, it was going to the `.dev` domain, whereas when doing `visit root_path` it was going to `127.0.0.1` and the cookies are not shared across domains, causing the user to appear to be logged out in the test.

I added additional test setup to set the default mailer host to be the Capybara host.

test/system/user_test.rb
bensheldon: Great idea! I added comments.

bensheldon:
@Mr0grog I addressed your feedback and this is ready for re-review 🙌

test/system/users_test.rb
```diff
@@ -0,0 +1,119 @@
...
+
+class UsersTest < ApplicationSystemTestCase
+  test "Invite, register, confirm, promote, demote, delete" do
+    admin = users(:admin_user)
+    viewer_email = 'user@example.com'
+
+    #
+    # Administrator sends a user invitation
+    #
+    Capybara.using_session('admin') do
```
Mr0grog: Not an issue; just curious: why the switch from symbols to strings?

test/system/users_test.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    #
+    # User is logged out from the deleted account and prevented from logging back in
+    #
+    Capybara.using_session('user') do
+      visit root_path
+      refute page.has_content?("Logged in as #{viewer_email}"), "User should NOT have an active session"
+
+      fill_in 'Email', with: viewer_email
```
Mr0grog: This is failing for me now! I think this used to have `click_on 'Login'` before it.

On a separate note, I just realized that this test is not getting run in CI. Can you add another step for it in `.circleci/config.yml` after the existing test step? https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/c2c3e57045c96b5a0adadd1e4d5b70dd08c98a73/.circleci/config.yml#L36-L41

I think the new step just needs to run `bundle exec rake test:system`, right?

bensheldon: I went with `bin/rails test:system test` which is the magical order according to https://stackoverflow.com/a/49830043/241735

test/system/users_test.rb
bensheldon: That was a holdover from debugging the session logout problem. I changed it back to symbols.

bensheldon:
@Mr0grog ok, I have the tests all passing. Let's give it another go 🏃

test/system/users_test.rb
Mr0grog: Huh, wonder what it is about Rake or the CLI parsing that makes the order so sensitive. Anyway, TIL!

PR #533, opened 2019-05-18
Add granular permissions system for users

bensheldon:
Contributes to #34.

- Adds individual permissions to "view", "import", "annotate" and "manage users". These are stored on the User inside of a postgres string array. 
- Migrates previous admins to "manage_users" and provides all existing users the "view", "import" and "annotate" permissions
- Installs the `pundit` gem to create access/authorization "policies" 

Permissions have been applied to:
- [x] Require `manage_users` permission to access admin pages
- [x] Require `view` permission to access any API data
- [x] Require `annotate` permission to access `api/annotations` actions
- [x] Require `import` permission to access `api/imports` actions

Questions:
- [x] Is there an API where I should expose the permissions of the current user? I'm not sure how the web frontend will want to consume them so that it can provide an appropriate UI.


Mr0grog:
> Is there an API where I should expose the permissions of the current user?

Externally (e.g. in [web-monitoring-ui](https://github.com/edgi-govdata-archiving/web-monitoring-ui)) we read information about the user from the Devise’s API at `/users/session`, which maps to [`app/controllers/users/sessions_controller.rb#validate_session`](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/89b016571b0275d53028b10b71cfd5d34117088e/app/controllers/users/sessions_controller.rb#L5-L16), which just renders the [`User` model](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/89b016571b0275d53028b10b71cfd5d34117088e/app/models/user.rb#L19-L26) as JSON. The same endpoints when accessed via POST generate JWT tokens for the UI to use, so it’s sort of a crappy halfway-to-oauth situation.

This is all definitely not really ideal in that it sits totally outside the normal `/api/v0` routes; it was slammed together early on with intention of being replaced by a separate auth system that handled all the old datarescue, archivers, etc. stuff, but things never really worked out there for a variety of reasons.

Mr0grog:
All that is to say: as long as you make sure the `User` model includes permissions when serializing to JSON, you are satisfying all the current API needs.

db/migrate/20190518154841_add_permissions_to_users.rb
```diff
@@ -0,0 +1,19 @@
+class AddPermissionsToUsers < ActiveRecord::Migration[5.2]
+  def up
+    add_column :users, :permissions, :string, array: true, default: []
+    add_index :users, :permissions, using: 'gin'
```
Mr0grog: My general tendency is to try and be conservative with indexes (maybe that’s wrong; it certainly makes it easier to shoot yourself in the foot later) — and it doesn’t seem like we would use this one. Do we need it?

db/migrate/20190518154841_add_permissions_to_users.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    execute "UPDATE users SET permissions = ARRAY['view', 'annotate', 'import'];"
+    execute "UPDATE users SET permissions = array_append(permissions, 'manage_users') WHERE admin = TRUE;"
+
+    remove_column :users, :admin
```
Mr0grog: Having been bitten before by problematic schema updates, I’d like to separate the addition of the `permissions` column from the removal of the `admin` column, and ship the former *before* the rest of this PR so that we don’t have any problems with the code and DB being out of sync during rolling updates or if we need to revert a change.

db/migrate/20190518154841_add_permissions_to_users.rb
bensheldon: Sounds good 👍 I removed them.

db/migrate/20190518154841_add_permissions_to_users.rb
bensheldon: 👍 I removed the removal of the `admin` column from the migration

bensheldon:
@Mr0grog This PR now has the full behavior of authorizing the API and is ready for review.

Mr0grog:
Sweet, thanks! Can’t promise I will get to it today, but will give it a solid read-through in the next few days.

test/controllers/api/v0/imports_controller_test.rb
```diff
@@ -19,24 +19,8 @@ def diff(_change, _options = nil)
...
-  end
-
-  def teardown
-    WebMock.disable_net_connect!
-    Archiver.allowed_hosts = @original_allowed_hosts
-  end
-
-  test 'can import data' do
-    import_data = [
+  def import_data
```
Mr0grog: Is there a reason you extracted this out, even though it’s only used in one function? All the other test functions still define their own `import_data` internally, since it’s mostly unique in every one.

app/views/admin/index.html.erb
```diff
@@ -4,10 +4,13 @@
       <th>ID</th>
       <th>E-mail</th>
       <th>Confirmed</th>
+      <th>View</th>
+      <th>Annotate</th>
+      <th>Import</th>
       <th>Admin</th>
```
Mr0grog: Oh, one more: we should probably drop this column :)

app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Admin
+  class UsersController < ApplicationController
```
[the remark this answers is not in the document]
bensheldon: ;-) I was thinking the same thing. I'll do it in a followup.

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,24 @@
...
   include SameOriginSessionsConcern
 
   before_action :set_environment
 
+  rescue_from Pundit::NotAuthorizedError, with: :not_authorized
+
   private
 
+  def not_authorized
+    return render(json: {}, status: :unauthorized) if request.xhr?
```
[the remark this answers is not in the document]
bensheldon: I'll make it `request.format == :json` for now.

Where my thoughts are going is to more cleanly separate out base `ApplicationController` for the HTML parts of the app and `ApiController` for the API parts and separately handle the authorization errors rather than having this one method cover both cases.

test/controllers/api/v0/imports_controller_test.rb
bensheldon: I'll inline it. I am struggling a bit with the mintest `test_` style because it seems like each one requires a lot of duplicated setup code. 

If you're ok with it, in future work I might explore minitest's spec syntax (`describe`, `it`) because it makes it a bit easier to create shared context across multiple tests.

app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    private
+
+    def user_params
+      params.require(:user).permit(permissions: []).tap do |filtered_params|
+        filtered_params[:permissions].delete('0') if filtered_params[:permissions]
+      end
+    end
+
+    def require_admin!
```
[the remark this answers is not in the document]
bensheldon: I removed it entirely, along with `admin?`

app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      else
+        render :edit
+      end
+    end
+
+    private
+
+    def user_params
+      params.require(:user).permit(permissions: []).tap do |filtered_params|
+        filtered_params[:permissions].delete('0') if filtered_params[:permissions]
```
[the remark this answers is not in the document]
bensheldon: Agreed about pain point. I added some docs to it. Personally I prefer Rails helpers paving over browser/html weirdness but agree that someones gotta discover it somehow.

I added a big comment block.

app/models/user.rb
```diff
@@ -25,6 +41,17 @@ def serializable_hash(options = {})
     super(options)
   end
 
+  def permission?(permission)
+    permissions.include? permission
+  end
+
+  PERMISSIONS.each do |permission|
+    define_method "#{permission}?" do
```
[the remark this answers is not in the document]
bensheldon: Love it. I went with `can_#{permission}?`

test/controllers/api/v0/imports_controller_test.rb
bensheldon: 👍

app/controllers/application_controller.rb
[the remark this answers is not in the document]
Mr0grog: That sounds like a good idea. 👍

test/controllers/api/v0/imports_controller_test.rb
Mr0grog: I am cool with that. I also have no objection if you want to migrate things over to RSpec, too.

IIRC, when I first started, I didn’t know Minitest had that syntax, and later when I discovered it, a bunch of things actually worked in a significantly different way (and I had trouble finding good resources on it), so I didn’t feel like it was worth the effort at the time. (Sadly, I never wrote any of that down and don’t remember details, so take that however you will.) And the use of Minitest in general was just because I didn’t think this project or the tests would get that big, and RSpec seemed like it might be an unnecessary large dependency. Obviously that hasn’t really held true.

PR #534, opened 2019-05-20
Use relative host for swagger; place actual file in app/assets

### LINKED ISSUE #292: Swagger form on home page always calls production instead of the server it is running on
## Steps to reproduce

1. Setup locally
1. Did partial import of seeds. (ran for 15 minutes, but had to abort, assuming there's be some data in the database?)
1. Logged in via credentials in seeds.rb: https://imgur.com/893yWlU
1. Tried to execute to test `GET /pages` via UI with default params

## Expected Behavior
Expected to see some response with data

## Current Behavior

Got: `401 Error: Unauthorized` message

## Your Environment
Running through Heroku backend services (see other PR), because my system has issue :)

Any thoughts?

### LINKED ISSUE #455: Symlink for swagger.yml in assets breaks in images built from git checkouts on Windows
A file containing the literal `../../../swagger.yaml` from the git symlink is returned instead of the linked contents of swagger.yaml

## Expected Behavior
I'd expect to get the actual swagger.yaml contents, so that swagger works.

## Current Behavior
Swagger doesn't work at all, because a valid config can't be loaded

## Possible Solution
Rather than symlinking, we could git ignore swagger.yml in assets and copy the file to the new location in the docker build.

## Steps to Reproduce (for bugs)
This is really not a bug, but rather a limitation of the way git works on windows.

## Context
This is an inconvenience for exploring the API, but not at all a show stopper. Really this is a note to self so that I don't forget.

## Your Environment
It's a unicorn:
- msys2 on windows 10
- minikube

Even though all of the building is happening in a linux environment via minikube, the git checkout is on windows. This is the source of the issue.

bensheldon:
- Removes hardcoded host and protocol which are [optional in the swagger spec](https://swagger.io/docs/specification/2-0/api-host-and-base-path/#host) to fix #292 
- Places the actual `swagger.yml` file in `app/assets` and places a symlink in the project root to fix #455

### DIFF
--- a/app/assets/data/swagger.yaml
+++ b/app/assets/data/swagger.yaml
@@ -9,7 +9,6 @@ info:
     name: GPL 3
     url: >-
       https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/LICENSE
-host: api.monitoring.envirodatagov.org
 basePath: /api/v0
 tags:
   - name: pages and versions
@@ -18,8 +17,6 @@ tags:
     description: comparisons between two Versions
   - name: imports
     description: Bulk-importing Versions
-schemes:
-  - https
 
 parameters:
   sort:
--- a/app/assets/data/swagger.yml
+++ b/app/assets/data/swagger.yml
@@ -1,1 +1,0 @@
-../../../swagger.yaml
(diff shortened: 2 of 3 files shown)


Mr0grog:
Deployed in https://github.com/edgi-govdata-archiving/web-monitoring-ops/commit/1a30d1145386fd7b081fa3f69cdabefe7f64c3fc

PR #535, opened 2019-05-21
Delete a newly created page if the version import fails

### LINKED ISSUE #192: An import that fails between creating a page and creating a version leaves an empty page record
If an import results in the creation of a new page record, but fails when creating the actual *version* of that page (because of invalid data), the empty page record gets left lying around afterward. While we think it’s OK for the DB to have pages with no versions, it’s not really desirable, and we shouldn’t wind up in that situation accidentally like this.

The import versions job needs to be careful to roll back the creation of a page if the version failed.

I can think of a couple straightforward ways of doing this:

- Wrap the page + version creation together in a single transaction. Not sure on the overhead of this.
- Keep a list of pages that were created by the job and, when cleaning up at the end, remove any that have no versions.

We also need a test for this.

bensheldon:
This fixes #192.

This PR is a broad wrapping of the import behavior with a rescue block because I'm not quite sure what the failure behavior is of importing. It's possible to target more specific import failures if you can point me in that direction. 

I also considered wrapping it in a transaction but I wanted to suggest the rescue block first. If I changed to a transaction, I'd want to do a bit of reordering in that method so I could open the transaction _after_ the API call so that the open transaction is as brief as possible.

### DIFF
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -106,7 +106,7 @@ class ImportVersionsJob < ApplicationJob
       @added << version unless existing
     rescue StandardError
       # Delete a newly created page if the version import fails
-      if page.uuid_previously_changed? && page.versions.empty?
+      if page.uuid_previously_changed? && page.versions.count.zero?
         page.maintainers = []
         page.destroy!
       end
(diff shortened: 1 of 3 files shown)


bensheldon:
@mr0grog no worries. I’ll close this for now.

PR #547, opened 2019-06-17
Update Rails binstubs

Mr0grog:
I was having trouble running `rails generate migration`, and it turns out it's because binstubs were out of date. This is basically just the result of running `bundle exec rake rails:update:bin`. I’m not sure what exactly changed about Spring usage here, but it seems to fix the issue.

### DIFF
--- a/bin/rails
+++ b/bin/rails
@@ -1,9 +1,4 @@
 #!/usr/bin/env ruby
-begin
-  load File.expand_path('../spring', __FILE__)
-rescue LoadError => e
-  raise unless e.message.include?('spring')
-end
 APP_PATH = File.expand_path('../config/application', __dir__)
 require_relative '../config/boot'
 require 'rails/commands'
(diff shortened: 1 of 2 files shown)


bin/rails
```diff
@@ -1,9 +1,4 @@
 #!/usr/bin/env ruby
-begin
-  load File.expand_path('../spring', __FILE__)
-rescue LoadError => e
-  raise unless e.message.include?('spring')
-end
```
bensheldon: I think this will remove the spring functionality entirely. Did you try running `bin/spring stop` when you were debugging spring?  I've seen spring (often) get into a weird state that requires restarting it.

Mr0grog: Hmmm, I was assuming it was getting loaded some other way in Rails now, otherwise this wouldn’t have been removed. My symptom was that `rails <anything>` would just hang, but I don’t think I thought to try stopping spring.

Mr0grog: Yeah, rolling this back, everything seems to be fine now, so I guess this was just a transient Spring issue. >:(

PR #548, opened 2019-06-17
Add indexes for commonly searched fields on Version

Mr0grog:
We frequently filter Versions by several fields that don't have indexes, and that causes some real slowdowns. This adds indexes for them.

In local development, this hasn’t been a problem, but our Versions table in production has gotten big enough that we are seeing some real slowdowns. There’s also probably other optimizations we could make here, but I am no Postgres expert and this is a good start for now.


Mr0grog:
This has been on hold because of some other distractions, so I’ve rebased it on top of current master.

Additionally, more database jiggery-pokery I’ve done lately has taught me that these indexes are not the principle issue (though they still matter). Instead, we have major hardware issues to fix and some re-thinking to do about how we design queries and about how we handle pagination.

PR #550, opened 2019-06-27
Give new users view and annotate permissions

### LINKED ISSUE #549: New users are not automatically getting “view” permissions
New users should be automatically granted “view” (and maybe “annotate”) permissions, but that’s not happening right now.

Mr0grog:
When users are first created, make sure they are granted `view` and `annotate` permissions.

Fixes #549.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -22,6 +22,7 @@ class User < ApplicationRecord
   has_one :invitation, class_name: 'Invitation', foreign_key: 'redeemer_id', inverse_of: :redeemer
 
   validates :permissions, contains_only: PERMISSIONS
+  before_create :set_defaults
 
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -59,4 +60,10 @@ def after_confirmation
       self.invitation = nil
     end
   end
+
+  def set_defaults
+    if self.permissions.empty?
+      self.permissions << VIEW_PERMISSION << ANNOTATE_PERMISSION
+    end
+  end
```
bensheldon: Give this a try; I think it's the most modern method: https://stackoverflow.com/a/43484863/241735

Mr0grog: Ooooh, sweet! Will use that.

Issue #191, opened 2018-01-05
Keep a record of what pages/versions an import touched and how


danielballan:
I have seen a (yet unresolved) bug where an errored import job causes the UI `/pages` endpoint to become un-renderable. For issues like this, it would be useful to be able to identify and potentially remove records generated by a specific Import.

Mr0grog:
How is it unrenderable??? Regardless of whatever craziness is happening in the importer, it shouldn’t be possible for `/api/v0/pages` to somehow choke on anything it’s pulling out of the DB. Is it raising an exception?

I guess I can see some value in linking a version/page to an import, but it doesn’t seem (to me, at least) like a great solution to the problem you’re describing.

danielballan:
> I guess I can see some value in linking a version/page to an import, but it doesn’t seem (to me, at least) like a great solution to the problem you’re describing.

So -- I definitely agree that we need to understand and fix the UI problem on its own. _Nothing_ we do with imports should ever be able to break the `/pages` view. It's a separate issue. It just highlighted for me that an import-batch-gone-wrong is something we're likely to see again.

danielballan:
See https://github.com/edgi-govdata-archiving/web-monitoring-ui/issues/174 for the UI bug.

Mr0grog:
See #192 for the DB bug related to specific issue that precipitated this.

Mr0grog:
Re-titled this to get at how @danielballan and I have talked about this issue. It doesn’t make sense to store information about what import created a page/version because multiple imports could modify a page or version (if you use `?update=merge` or `?update=replace`). There’s also a lot of overhead here.

Instead, we probably want to log what each import did. Some possible approaches here:

1. Store a structured output log alongside the stored import payload, where each row in the import payload has a corresponding row that is something like:

    ```js
    {
      "page_uuid": "<uuid>",
      "page_operation": "<created|updated|null>",
      "version_uuid": "<uuid>",
      "version_operation": "<created|updated|null>",
      "at": "2018-03-01T00:01:52Z"
    }
    ```

2. Store a structured log of operations alongside the stored import payload, where rows don’t necessarily correspond to the payload, but are just a list of operations, like:

    ```js
    {"object": "page", "operation": "created", "uuid": "<uuid>", "at": "2018-03-01T00:01:52Z"}
    {"object": "version", "operation": "created", "uuid": "<uuid>", "at": "2018-03-01T00:01:52Z"}
[...]

PR #551, opened 2019-06-27
Log all operations an import does

bensheldon:
Connects to #191.

This logs imported pages and versions. It uses the file attachment tool to attach ndjson-ified log objects. Some notes:

- I struggled a bit clearly naming the database column and methods. Advice appreciated
- I'm not sure if the log lifecycle is sufficient. It will attempt to append to an existing logfile if it exists (implying that an import might be run multiple times).
- I didn't create any accessors for the logs (other than downloading the file `#load_file`), assuming that they would only be accessed in another context than they're being written to (e.g. importer writes the logs, the UI/console reads them). I can though if that'd be helpful.
- I feel like the array slicing/joining is a bit repetitive, though I played with some refactorings and this was the best I could do for now. 

Todo:
- [x] Add the row associated with the log
- [x] Add log lines marking the start and end of the import job, e.g. {"object": "import", "operation": "start", "at": "2019-06-10T01:03:15Z"}
- [x] @import.update_behavior can be :skip, :replace, or :merge:
- [x] Use `import-{id}.log` for filename
- [x] Remove `uuid_previously_changed?` usage
- [x] Include why a version was skipped
- [x] Cache log contents so less fetching from s3

### DIFF
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -102,6 +102,7 @@ class ImportVersionsJob < ApplicationJob
     version.update_different_attribute(save: false)
     version.save
 
+    log(object: version, operation: :created)
     @added << version unless existing
   end
 
@@ -151,6 +152,12 @@ class ImportVersionsJob < ApplicationJob
     page.add_maintainer(record['site_agency']) if record.key?('site_agency')
     (record['page_tags'] || []).each {|name| page.add_tag(name)}
     page.add_tag("site:#{record['site_name']}") if record.key?('site_name')
+    
+    if page.uuid_previously_changed?
+      log(object: page, operation: :created)
+    else
+      log(object: page, operation: :updated)
+    end
 
     page
   end
@@ -160,6 +167,15 @@ class ImportVersionsJob < ApplicationJob
   def warn(message)
     @import.processing_warnings << message
     Rails.logger.warn "Import #{@import.id} #{message}"
+  end
+
+  def log(object:, operation:)
+    @import.add_log(
+      uuid: object.uuid,
+      object: object.class.name.parameterize,
+      operation: operation,
+      at: object.updated_at
+    )
   end
 
   # iterate through a JSON array or series of newline-delimited JSON objects
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -32,7 +32,7 @@ class ImportVersionsJob < ApplicationJob
     each_json_line(raw_data) do |record, row, row_count|
       begin
         Rails.logger.info("Importing row #{row}/#{row_count}...") if Rails.env.development? && (row % 25).zero?
-        import_record(record)
+        import_record(record, row)
       rescue Api::ApiError => error
         @import.processing_errors << "Row #{row}: #{error.message}"
       rescue ActiveModel::ValidationError => error
@@ -58,7 +58,7 @@ class ImportVersionsJob < ApplicationJob
     end
   end
 
-  def import_record(record)
+  def import_record(record, row)
     page = page_for_record(record, create: @import.create_pages)
     unless page
       warn "Skipped unknown URL: #{record['page_url']}@#{record['capture_time']}"
@@ -67,6 +67,12 @@ class ImportVersionsJob < ApplicationJob
     unless page.active?
       warn "Skipped inactive URL: #{page.url}"
       return
+    end
+
+    if page.uuid_previously_changed?
+      log(object: page, operation: :created, row: row)
+    else
+      log(object: page, operation: :found, row: row, at: Time.current) # TODO: is it ok to use :found because the page is not updated
     end
 
     existing = page.versions.find_by(
@@ -102,7 +108,7 @@ class ImportVersionsJob < ApplicationJob
     version.update_different_attribute(save: false)
     version.save
 
-    log(object: version, operation: :created)
+    log(object: version, operation: :created, row: row)
     @added << version unless existing
   end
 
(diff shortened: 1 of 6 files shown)


Mr0grog:
@bensheldon just checking in — still working on this?

bensheldon:
@Mr0grog yep! slowly 🐢

app/models/import.rb
```diff
@@ -19,10 +20,41 @@ def load_data
...
+
+  def persist_logs
+    return if unpersisted_logs.empty?
+
+    if log_file.present?
+      existing_logs = load_logs
+      existing_logs << "\n" + unpersisted_logs.join("\n")
+      FileStorage.default.save_file(log_file, existing_logs)
+    else
+      self.log_file = SecureRandom.uuid
```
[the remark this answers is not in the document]
bensheldon: It doesn't appear that the filestorage implementation supports subdirectories. I received this error when trying to use one; I think it's not noticing/creating a subdirectory on the filesystem (probably would work on S3 though):

`Errno::ENOENT: No such file or directory @ rb_sysopen - /Users/bensheldon/Repositories/edgi-govdata-archiving/web-monitoring-db/tmp/storage/working/import_logs/import-67.log`

bensheldon:
@Mr0grog I completed the 3 items that you identified (summarized in checkmarks in the PR summary above). Please take another look and let me know what other changes you need.

[I'm never sure if I should comment on your review comments or what other than to say "I did this"]

app/jobs/import_versions_job.rb
```diff
@@ -69,6 +71,12 @@ def import_record(record)
       return
     end
 
+    if page.uuid_previously_changed?
+      log(object: page, operation: :created, row: row)
```
Mr0grog: How come this isn’t being done down in `page_for_record`? I had to spend a minute puzzling out why you were checking `uuid_previously_changed?` (it was non-obvious to me that what’s really happening is that the uuid would only have changed if the object was new), plus it seems prone to potential breakage if `page_for_record` changes.

app/jobs/import_versions_job.rb
```diff
@@ -69,6 +71,12 @@ def import_record(record)
...
+    else
+      log(object: page, operation: :found, row: row, at: Time.current) # TODO: is it ok to use :found because the page is not updated
```
Mr0grog: I think it’s OK not to log this situation — my intuition is that it would be more noisy than helpful when trying to decipher what led to a failure from the logs. Do you think it’s critical information to keep around or would be especially helpful?

app/jobs/import_versions_job.rb
```diff
@@ -102,6 +111,13 @@ def import_record(record)
     version.update_different_attribute(save: false)
     version.save
 
+    if version.uuid_previously_changed?
```
Mr0grog: You should just be able to check `existing` here, which would be a bit more clear. Or am I missing something?

```suggestion
    unless existing
```

app/jobs/import_versions_job.rb
```diff
@@ -94,6 +102,7 @@ def import_record(record)
     end
 
     if @import.skip_unchanged_versions? && version_changed?(version)
+      log(object: version, operation: :skipped, row: row)
```
Mr0grog: Might be helpful to include something clarifying here that we skipped it because it was identical to the version preceding it, rather than because we already have a copy of it (that skip happens up on line 85).

app/jobs/import_versions_job.rb
```diff
@@ -102,6 +111,13 @@ def import_record(record)
     version.update_different_attribute(save: false)
     version.save
 
+    if version.uuid_previously_changed?
+      log(object: version, operation: :created, row: row)
+    else
+      log_version_operation = { 'merge' => 'merged', 'replace' => 'replaced' }.fetch(@import.update_behavior) { @import.update_behavior }
```
Mr0grog: Minor nit: can you pull the mapping object out into the class here so we don't allocate a whole bunch of these? Or just use `@import.update_behavior` directly (being able to map to a specific value for the logs is a nice touch, though).

app/models/import.rb
```diff
@@ -19,10 +20,41 @@ def load_data
     FileStorage.default.get_file(file)
   end
 
+  def load_logs
+    return if log_file.blank?
+
+    FileStorage.default.get_file(log_file)
```
Mr0grog: Would it be a good idea to cache this for the next call? Since we save many times in the middle of an import, this could slow things down a lot. (Sorry I must have missed this the last time around.)

app/models/import.rb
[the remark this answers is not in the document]
Mr0grog: Oh, `LocalFile` doesn’t attempt to create subdirectories (only the main directory for the storage) before writing. Hmmm, do you think that’s worth fixing?

Not worth taking extra effort on it unless you want to; the new naming should make it possible to sort things out without directories.

app/models/import.rb
Mr0grog: Alternatively, we could load + write the logs on a thread so they don’t block.

app/jobs/import_versions_job.rb
bensheldon: I honestly don't really know what the shape of behavior is for imports (e.g. what the ratio is to pages and versions and whether it's more likely for a page to be newly created or already existing). 

But, from purely reading the complexity of code around whether to find or create a new page, it seems extra important to me to note what happened :-)

app/jobs/import_versions_job.rb
bensheldon: My original intention was to keep all of the logging in the `#import_record` method, both for conceptual reasons  and also so I didn't have to pass down the `row` variable into every other `#*_for_record` method.  ...though I think in practice the original concept wasn't practical once I integrated your last round of feedback.

I'll drop it down into the extracted `page_for_record`.

app/jobs/import_versions_job.rb
bensheldon: ✅ 

I also renamed `existing` to be `existing_version`

PR #577, opened 2019-08-09
Check `different` flag in `version.next/previous` and associated methods

### LINKED ISSUE #576: Versions ?include_change_from_* and various `previous`/`next` etc. should check `different`
One of the analysts recently pointed out a situation where they weren’t getting the correct links in their task sheet. Specifically, the sheet should link to the diff from the most recent version *before* the current time period to the latest version in the time period:

```
/pages/<page id>/<version before period>..<latest in period>
```

But instead, they were just getting a comparison from the second-latest in the period to the latest some of the time:

```
/pages/<page id>/..<latest in period>
```

This happens whenever we can’t figure out the what the latest version before the time period was. We get that information by examining the `change_from_previous` field of the first version in the period (`change_from_previous.uuid_from` indicates the latest version before). There are two major cases where this fails:

1. Analysis failed (or didn't run because the content type was not analyzable). This isn’t that common (maybe one of several thousand fails unless we have some bigger systems failure).
2. New, non-different versions were inserted after doing the analysis, and we never analyzed the new change from previous because it's not different so there's no point.

It turns out (2) is happening a lot. It looks to me like the bigger problem behind that is really that when we ask for the `previous` and `next` version or change, we don’t check the `different` flag: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/7c009472103970909e0fe0b11066322298a22911/app/models/version.rb#L21-L35

I think the right thing is to make these methods check `different=true` by default (and take an argument to turn that check off).

Mr0grog:
When looking for the `next` or `previous` version, we should almost always be looking for the next/previous version that was *different*. This was causing a lot of issues with calculating links for analysts, because new, non-different versions would be inserted after an analysis was done, and the analysis would be lost (because we'd try and retrieve it for the current vs. "previous" version, but "previous" was no longer the right thing).

This makes the default behavior one in which we check the different flag, but optionally lets you say `next(different: false)` to get the next version regardless of whether it's different.

This will also have the happy byproduct of making sure we are always analyzing versions that are different—it looks like we may have frequently been doing relatively pointless analyses before.

Fixes #576.

### DIFF
--- a/app/models/version.rb
+++ b/app/models/version.rb
@@ -18,32 +18,38 @@ class Version < ApplicationRecord
     self.page.versions.reorder(capture_time: :asc).first
   end
 
-  def previous
-    self.page.versions.where('capture_time < ?', self.capture_time).first
+  def previous(different: true)
+    self.page.versions
+      .where('capture_time < ?', self.capture_time)
+      .where(different: different)
+      .first
   end
 
-  def next
-    self.page.versions.where('capture_time > ?', self.capture_time).last
+  def next(different: true)
+    self.page.versions
+      .where('capture_time > ?', self.capture_time)
+      .where(different: different)
+      .last
   end
 
-  def change_from_previous
-    Change.between(from: previous, to: self, create: nil)
+  def change_from_previous(different: true)
+    Change.between(from: previous(different: different), to: self, create: nil)
   end
 
-  def change_from_next
-    Change.between(from: self, to: self.next, create: nil)
+  def change_from_next(different: true)
+    Change.between(from: self, to: self.next(different: different), create: nil)
   end
 
   def change_from_earliest
     Change.between(from: earliest, to: self, create: nil)
   end
 
-  def ensure_change_from_previous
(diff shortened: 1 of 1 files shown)


Mr0grog:
Going to go ahead and merge & deploy this tomorrow morning, since it seems to work well locally, passes tests, and we need time to re-run a possible week’s worth of analyses before the next analyst spreadsheet run (happens Tuesday mornings at midnight ET).

Mr0grog:
(Unless someone wants to take a look and finds something concerning, obviously!)

Mr0grog:
> I think you could implement these methods as Association Extensions instead of methods on the model instance.

Are you thinking specifically of the `change_from_<whatever>` methods? So we’d have something like `version.changes.from_previous`? Or am I missing what you’re getting at?

In any case, I think I’m going to merge this as-is so we have it working for analysts ASAP, but am curious to understand more about how to better structure it.

Mr0grog:
(Update: deployed this and scheduled 26,437 version analyses.)

Issue #348, opened 2018-07-27
Remove “lightweight query” from pages controller


Mr0grog:
The `PagesController` has a ridiculous custom query implementation called `lightweight_query` that loads data from the DB without instantiating ActiveRecord models. It saves a lot of memory and runs much faster when the queries are large (i.e. when using the `?include_versions` argument to include every version alongside each page in the response), but:

- We are slowly removing the `?include_versions` query arg that makes this optimization important
- We have since upgraded to much beefier machines, so this optimization isn’t as valuable anymore
- It is a serious pain in the butt to maintain (changes to the `PagesController` are hard to do well; Rails upgrades require a lot more testing than they should, e.g. #286).

We should probably #310 first, but then we really should do this.

See the code in question here: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/2619d56637a7dcfb1525308da12020628a7e113c/app/controllers/api/v0/pages_controller.rb#L174-L309

Mr0grog:
We turned off the functionality in #582. It’s been running well for a little over two weeks now, so it’s probably time to completely excise it from the code.

PR #582, opened 2019-08-21
Switch off `lightweight_query` by default

Mr0grog:
This is a first step to #348, removing `lightweight_query`. It adds an environment variable, `USE_LIGHTWEIGHT_QUERY`. If the variable is set, the old functionality will be used, but if not, it will default to normal ActiveRecord functionality. This way, we can have an emergency switch to flip it back on if things go wrong. In a later PR, we can remove the functionality (and this env var) entirely.

Note this has a kind of complicated way of accomplishing the `?include_versions` query. I couldn't figure out how to make it work in ActiveRecord as nicely as it does in `lightweight_query`, but I'm also not *super* worried about it since that whole thing is turned off and will probably be removed anyway. @bensheldon if you can make this work better than me, go for it.

### DIFF
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -51,9 +51,10 @@ class Api::V0::PagesController < Api::V0::ApiController
             .where(uuid: page_ids)
             .includes(:tags, :maintainers)
             .order(sorting_params.present? ? sorting_params : 'pages.updated_at DESC')
+            .index_by(&:uuid)
           # Join them up!
-          results.each_with_index do |page, index|
-            other_page = additions[index]
+          results.each do |page|
+            other_page = additions[page['uuid']]
             page['tags'] = other_page.taggings.as_json
             page['maintainers'] = other_page.maintainerships.as_json
           end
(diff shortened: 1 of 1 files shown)


Mr0grog:
On my local machine, queries vary from about the same to about half as fast as they were with `lightweight_query`. A relatively similar time is spent in Postgres, and most of the extra time is other Ruby code chugging away.

app/controllers/api/v0/pages_controller.rb
```diff
@@ -26,13 +26,41 @@ def index
...
+            .order('versions.capture_time DESC')
+            .as_json(include: :versions)
+          # Tags and Maintainers
+          additions = Page
+            .where(uuid: page_ids)
+            .includes(:tags, :maintainers)
+            .order(sorting_params.present? ? sorting_params : 'pages.updated_at DESC')
+          # Join them up!
+          results.each_with_index do |page, index|
+            other_page = additions[index]
```
bensheldon: [Suggestion] The combination of the two queries is the only thing that makes me (slightly) pause because I can think of vague situations where a numerical index could get out of sync between the two queries. I think it might be slightly safer, and not much less performant, to use a lookup hash keyed on the uuid using  `index_by`:


```ruby
additions  = Page.where...index_by(&:uuid)

results.each do |page|
  other_page = additions[page.uuid]
  ...
end
```

Mr0grog: Ooh, nice, I did not know about `index_by`!

PR #595, opened 2019-09-04
Document `ALLOWED_ARCHIVE_HOSTS` and S3 Buckets

### LINKED ISSUE #249: Document how archiving and `ALLOWED_ARCHIVE_HOSTS` work

### LINKED ISSUE #469: Document details of S3 buckets
Add S3 bucket details to documentation from https://github.com/edgi-govdata-archiving/web-monitoring/issues/108 

At current, we have three S3 buckets:

1. An archive bucket, which is publicly viewable and controlled by a `web-monitoring-db `server. When a new version is imported and its `uri` field does not match one of the hosts listed in the `ALLOWED_ARCHIVE_HOSTS` setting, it gets stored here for safe keeping and public access.
2. A working bucket, which is not publicly viewable and is controlled by a `web-monitoring-db `server. It stores things like raw import lists, temp files, etc.
3. A versionista bucket, which is publicly viewable and controlled by the Versionista scraper and actually has pretty much nothing at all to do with w`eb-monitoring-db`. (The URL of this bucket is listed in `ALLOWED_ARCHIVE_HOSTS` [see (1) above], so it never really gets touched by DB.)

Depending on how detailed you want these diagrams to be, those might or might not be worth differentiating — too much detail could be confusing. (But we do need to clarify them somewhere, probably in -`db’`s docs.)

Mr0grog:
Add more explanation to `.env.example` and to the README about how data is managed and stored in the application. Also explain the overall data model and document the public access user on staging (see edgi-govdata-archiving/web-monitoring-ui#220).

Hopefully this is clarifying and not confusing! See here for a full render of the new README instead of a diff, if that helps: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/249-469-explain-just-who-gets-trusted-to-hold-your-gooses-golden-eggs-and-where/README.md

Fixes #249.
Fixes #469.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,22 @@
 
 # web-monitoring-db
 
-This repository is the database and API underlying the EDGI [Web Monitoring Project](https://github.com/edgi-govdata-archiving/web-monitoring).
+This repository is the database and API underlying the EDGI [Web Monitoring Project](https://github.com/edgi-govdata-archiving/web-monitoring). It’s a Rails app that:
 
-It’s a Rails app that:
+- Acts as a database of monitored pages and captured versions of those pages over time.
 
-- Acts as a database of monitored pages and revisions that have been made to them
-- Allows other services to add new tracked pages/versions (we are currently focused on Versionista, but this database will soon host data from other sources, such as the Internet Archive)
-- Provides an API to get that version data and allow analysts or other automated tools to annotate those versions with metadata
+    *(The application does not record new versions itself, but relies on importing data from external services, like [the Internet Archive](https://archive.org) or [Versionista](https://versionista.com). See [“How Data Gets Loaded”](#how-data-gets-loaded) below for more.)*
+
+- Provides an API to get that page and version data, and to allow analysts or other automated tools to annotate those versions with metadata about what has changed from version to version.
+
+For more about how data is modeled in this project, see [“Data Model”](#data-model) below.
+
+API documentation is available from the homepage of the application, e.g. by pointing your browser to http://localhost:3000/ or https://api.monitoring.envirodatagov.org. It’s generated from our OpenAPI docs in [`swagger.yml`](./swagger.yml).
+
+We maintain a publicly available *staging server* at https://api-staging.monitoring.envirodatagov.org that you can test against. It runs the latest code and has non-production data — it’s safe to modify or post new versions or annotations to, but you should not rely on that data sticking around; it may get reset at any time. **For access, ask for an account on Slack or use the public user credentials:**
+
+- Username: `public.access@envirodatagov.org`
+- Password: `PUBLIC_ACCESS`
 
 
 ## Installation
@@ -175,7 +184,7 @@ It’s a Rails app that:
     - `analysis`: Auto-analyze changes between versions and create annotations with the results.
 
 
-## Manual Postgres Setup
+### Manual Postgres Setup
 
 If you don’t want to populate your DB with seed data, want to manage creation of the database yourself, or otherwise manually do database setup, run any of the following commands as desired instead of `rake db:setup`:
 
@@ -197,7 +206,7 @@ User.create(
 ```
 
 
-## Docker
+### Docker
 
 The Dockerfile runs the rails server on port 3000 in the container. To build
 and run:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -212,6 +221,55 @@ docker run -p 6379:6379 envirodgi/db-import-worker -e <ENVIRONMENT VARIABLES> .
...
+
+Most of the data sent to `/api/v0/imports` matches up directly with the structure of the [`Version` model](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/models/version.rb). However, the `uri` field in an import is treated specially. If the `uri` host matches one of the hosts listed in the [`ALLOWED_ARCHIVE_HOSTS` environment variable](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/.env.example), the application simply stores that as the version’s `uri`. If it doesn’t match, the application downloads the content from `uri` and stores it in its `FileStorage`. The intent is to make sure data winds up at a reliably available location, ensuring that anyone who can access the API can also access the raw response body for any version. The application’s storage area can be the local disk or it can be S3, depending on configuration. The component can take pluggable configurations, so we can support other storage types or locations in the future.
```
lightandluck: The description about `uri` and what is stored is confusing here. When it says "the application simply stores that as the version's `uri`", I'm not sure what "that" is. It makes me think that some blob is stored in `uri` (which I'm assuming should be a string, but now I'm not sure, so I'm lost). 

When the hosts match, nothing extra is stored and the app just grabs from the archive host? Is that correct? If so, using the term "stored" may also be why it sounds confusing.

The explanation in the comments in `.env.example` are actually really clear! If we can transfer some of that language to this section, I feel it would be more understandable. 

I pulled some of the comments into the suggestion below as an example, but I'm not sure about verbiage and left some question marks. Please don't take the suggestion as-is!
```suggestion
Most of the data sent to `/api/v0/imports` matches up directly with the structure of the [`Version` model](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/app/models/version.rb). However, the `uri` field in an import is treated specially. 

When new page or version data is imported, the `uri` field points to a location where the raw HTTP response body is stored. If the `uri` host does *not* match one of the values in `ALLOWED_ARCHIVE_HOSTS`, the application downloads the data from `uri` and stores it in its `FileStorage`. If the `uri` host matches one of the hosts listed in the [`ALLOWED_ARCHIVE_HOSTS` environment variable](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/master/.env.example), the application simply [points to]?? [what was given]?? as the version’s `uri`. 

The intent is to make sure data winds up at a reliably available location, ensuring that anyone who can access the API can also access the raw response body for any version. The application’s storage area can be the local disk or it can be S3, depending on configuration. The component can take pluggable configurations, so we can support other storage types or locations in the future.
```

README.md
```diff
@@ -212,6 +221,55 @@ docker run -p 6379:6379 envirodgi/db-import-worker -e <ENVIRONMENT VARIABLES> .
...
+- **Maintainers**, which can be applied to pages. They represent organizations that maintain a given page. For example, the page at `https://www.epa.gov/citizen-science` is maintained by `EPA`.
+
+- **Imports** model requests to import new data and the results of the import operation.
+
+- **Users** model people (both human and bots) who can view, import, and annotate data. You currently have to have a user account to do anything in  the application, though we hope accounts will not be needed to view public data in the future.
+
+
+### How Data Gets Loaded
+
+The web-monitoring-db project does not actually monitor or scrape actual pages on the web. Instead, we rely on importing data from other services, like [the Internet Archive](https://archive.org). Each day, a script queries other services for historical snapshots and sends the results to the `/api/v0/imports` endpoint.
```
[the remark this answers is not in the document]
Mr0grog: In this case, I think it’s probably not—scraper is the wrong repo for this now. -processing has all the code for pulling stuff from Wayback; scraper only does versionista + generating analyst task sheets, which really shouldn’t be its responsibility…

Anyway, that part could be done in a lot of different ways in someone’s setup and will probably continue to be done in different ways in our own, and I’m more than a little bit worried about being *too* specific here and winding up with this getting way out of sync from reality.

README.md
Mr0grog: > When the hosts match, nothing extra is stored and the app just grabs from the archive host?

If they match, the app grabs nothing. It’ll just point to that external location whenever someone wants the raw response data for that version record.

Mr0grog: Ok, took another crack at this in 712782c91740d57118fc67ec0514ea9e0bd2e021. How does that read? https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/249-469-explain-just-who-gets-trusted-to-hold-your-gooses-golden-eggs-and-where/README.md#how-data-gets-loaded

README.md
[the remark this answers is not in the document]
lightandluck: Understood 👍

README.md
```diff
@@ -212,6 +221,64 @@ docker run -p 6379:6379 envirodgi/db-import-worker -e <ENVIRONMENT VARIABLES> .
...
+- **Users** model people (both human and bots) who can view, import, and annotate data. You currently have to have a user account to do anything in the application, though we hope accounts will not be needed to view public data in the future.
+
+Actual database schemas for each of these tables is listed in [`db/schema.rb`](./db/schema.rb).
+
+
+### How Data Gets Loaded
+
+The web-monitoring-db project does not actually monitor or scrape pages on the web. Instead, we rely on importing data from other services, like [the Internet Archive](https://archive.org). Each day, a script queries other services for historical snapshots and sends the results to the `/api/v0/imports` endpoint.
+
+Most of the data sent to `/api/v0/imports` matches up directly with the structure of the [`Version` model](./db/schema.rb). However, the `uri` field in an import is treated specially.
```
lightandluck: `Version` model was linking to `.../models/version.rb` before, but now points to `schema.rb`. It makes it clearer (for me at least!) what the table structure of `versions` is by looking at `schema.rb`, but not sure if `version.rb` is more useful for experienced Rails developers. Just double-checking this change was intentional

README.md
lightandluck: Looks good!

README.md
Mr0grog: > not sure if version.rb is more useful for experienced Rails developers.

It’s not! :)

I changed it because I realized while updating that this is all about what data is available from an object, and ActiveRecord classes don’t give you any of that information. It’s all in `schema.rb`. (The classes have extra methods and validations and so on, but the actual data properties are determined dynamically at runtime when data is loaded from a database query.)

Issue #579, opened 2019-08-20
Rethink Pagination


Mr0grog:
Last week, I made a bunch of changes to our database’s configuration (edgi-govdata-archiving/web-monitoring-ops#26) and to the indexes for the versions table (#548) to address massive performance issues we were hitting. Those have made a big difference, but we have one remaining major problem: the `count(*)` queries we use for pagination.

Most of our endpoints that return lists (e.g. `/api/v0/versions`) return a json:api inspired structure like:

```js
{
  "links": {
    "first": "https://api.monitoring.envirodatagov.org/api/v0/versions?chunk=1&chunk_size=100",
    "last": "https://api.monitoring.envirodatagov.org/api/v0/versions?chunk=23042&chunk_size=100",
    "prev": null,
    "next": "https://api.monitoring.envirodatagov.org/api/v0/versions?chunk=2&chunk_size=100"
  },
  "meta": {
    "total_results": 2304117
  },
  "data":[...]
}
```

In order to figure out the `last` page, the first thing we always do is run a `count(*)` version of the query that would produce our results (e.g. for `SELECT * FROM versions;`, we’d first do `SELECT count(*) FROM versions;`). That lets us calculate how many chunks/pages of results we have. As a bonus, we added the `meta.total_results` field, which several active use cases (e.g. the [healthcheck script](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/ce0d2e3d4c712b22196bf10da0a46a82fddf46d1/scripts/ia_healthcheck#L23-L36)) now depend on.

Unfortunately, `count(*)` is pretty slow in Postgres, especially if the whole table does not fit in memory. For example, `GET /api/v0/versions` currently takes a little over 18 seconds: 18 seconds for the count, and 30 *milliseconds* for the actual data. While use cases like the healthcheck require some kind of support for `count`, **we need to take `count` out of the critical path for pagination.**

Some ideas:

1. Do some fancy footwork and cache the `count` results, either in a table or in Redis. Every time we import, we would clear the cache.

2. Postgres 9.5+ has a fancy-pants [sampling system](https://www.postgresql.org/docs/devel/sql-select.html#SQL-FROM) for doing quick estimations: `SELECT 100 * count(*) AS estimated_total FROM versions TABLESAMPLE SYSTEM (1);` We could use estimated counts. (This still gets slower as the table grows, but relatively speaking, it’s quite speedy.)

[...]

PR #596, opened 2019-09-04
Only include totals in list endpoints if `?include_total=true`

Mr0grog:
For large tables, counting the number of possible results (in order to determine the URL for the *last* chunk in a result set) turns out to be extremly expensive. (For example, in our production `versions` table, the main query takes < 20 milliseconds, but counting the totals takes 18,000 milliseconds.) Instead, only count the total results and only include the URL for the last chunk if a user explicitly requests it.

In the future, we might move to an entirely different paging structure (that's not offset-based). See #579 for more.

NOTE: we need to update and redeploy scripts that use the `meta.total_results` before merging and deploying this:

- [x] [IA healthcheck](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/master/scripts/ia_healthcheck) in web-monitoring-processing
- [x] [Analyst task sheet generation](https://github.com/edgi-govdata-archiving/web-monitoring-versionista-scraper/blob/master/bin/query-db-and-email) in web-monitoring-versionista-scraper (edgi-govdata-archiving/web-monitoring-versionista-scraper#237)
- [x] Update documentation

### DIFF
--- a/app/assets/data/swagger.yaml
+++ b/app/assets/data/swagger.yaml
@@ -30,6 +30,38 @@ parameters:
       `title:asc,url:desc`. You can sort by any field on the result
       objects.
 
+  chunk:
+    name: chunk
+    in: query
+    description: pagination parameter
+    required: false
+    type: integer
+    default: 1
+
+  chunk_size:
+    name: chunk_size
+    in: query
+    description: number of items per chunk
+    required: false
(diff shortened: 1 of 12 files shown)


Mr0grog:
Merging this in, since no feedback. 🍿 💣

PR #613, opened 2019-10-08
Switched maintainer to label directive

### LINKED ISSUE #583: Switch MAINTAINER to LABEL directive in Docker
We use the `MAINTAINER` directive in our docker files, but it’s been deprecated:

> The `LABEL` instruction is a much more flexible version of this and you should use it instead, as it enables setting any metadata you require, and can be viewed easily, for example with docker inspect. To set a label corresponding to the `MAINTAINER` field you could use:
> 
> ```
> LABEL maintainer="SvenDowideit@home.org.au"
> ```

- [`MAINTAINER` docs](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated)
- [`LABEL` docs](https://docs.docker.com/engine/reference/builder/#label)

amoury:
closes #583 

Since `MAINTAINER` instruction is deprecated, the directive has been switched to `LABEL` which is a more flexible version and is recommended as per docs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 
 ### BASE ENVIRONMENT STAGE ###
 FROM ruby:2.6.3-slim as base
-MAINTAINER enviroDGI@gmail.com
+LABEL maintainer="enviroDGI@gmail.com"
 
 # Install apt based dependencies required to run Rails as
 # well as RubyGems. As the Ruby image itself is based on a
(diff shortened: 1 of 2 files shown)


amoury:
> Looks great, thanks! 👍
> 
> Would you like to be added to the [contributors list](https://github.com/edgi-govdata-archiving/web-monitoring-db#contributors)?

Yeah sure...Thank you 😊

Mr0grog:
Done! Thanks again for contributing. :)

Issue #594, opened 2019-09-03
Automatically tag news/blog pages with `news` tag


Mr0grog:
A while back, we manually tagged all pages with `news` or `blog` in the URL’s path using the `news` tag. We should probably just build this feature in for imports, the same way we automatically tag `domain:<domain>` and `2l-domain:<second-level domain>`.

You can see where we currently do this in the `Page` model: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/42bb8c7020e4d139fa278a95e0947cf9c8329659/app/models/page.rb#L122-L125

Tag a page as `news` if:

- The URL path contains `news`
- The URL path contains `blog`
- The URL path contains `press`
- ?

ewelinasobora:
I'd like to work on this issue :)

Mr0grog:
@ewelinasobora great! Let me know if you have any questions, or post a draft PR if you need any help or feedback before it’s entirely ready.

ewelinasobora:
@Mr0grog Thanks! Feedback will be helpful for sure :)

PR #618, opened 2019-10-13
Add news tag to pages with news, press, blog in a url

ewelinasobora:
It should resolve issue #594

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -38,7 +38,7 @@ class Page < ApplicationRecord
   has_many :maintainers, through: :maintainerships
 
   before_create :ensure_url_key
-  after_create :ensure_domain_tags
+  after_create :ensure_domain_and_news_tags
   before_save :normalize_url
   validate :url_must_have_domain
   validates :status,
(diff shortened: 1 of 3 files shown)


app/models/page.rb
```diff
@@ -124,8 +125,16 @@ def ensure_domain_tags
     self.add_tag("2l-domain:#{second_level_domain}")
   end
 
+  def ensure_news_tags
+    self.add_tag('news') if news
+  end
+
   protected
 
+  def news
```
Mr0grog: It would be a little more ruby-esque to name a boolean method like this with a question mark:

```suggestion
  def news?
```

app/models/page.rb
```diff
@@ -124,8 +125,16 @@ def ensure_domain_tags
     self.add_tag("2l-domain:#{second_level_domain}")
   end
 
+  def ensure_news_tags
```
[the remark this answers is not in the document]
ewelinasobora: fixed.

app/models/page.rb
ewelinasobora: fixed.

Issue #451, opened 2018-12-03
Add `status` field to Pages


Mr0grog:
We need to add a `status` field to page objects that attempts to represent what the overall status is for a page right now (we need this on a page level instead of just checking the latest version to account for things like intermittent errors). See edgi-govdata-archiving/web-monitoring#122.

Its value should be calculated on a moving time window. We think the right time window is 14 days, but this should be changeable (either a constant we can modify or a config variable we read in from the environment).

To calculate it, load all the versions of a page over the time window and:

1. If the status code is *successful* (i.e. < 400) more than X% of the time, mark the page’s `status` as `200`. (X should be configurable, but we think it’ll be somewhere between 50-80%.)

2. Otherwise, set the page’s `status` to the most common error status code (i.e. >= 400).

What needs to happen:

- [x] Add an int `status` field to the page record.
- [x] Add a method of updating it (this could be a job that we run regularly or it could be part of the import job). The actual calculation should probably be a method on the `Page` model. (See #619)
- [x] Maybe make this field queryable in the API?

This depends on #450.

danielballan:
I endorse!

Mr0grog:
This work is halfway done. I definitely don’t want to leave it in that state.

Mr0grog:
The final work on scheduling automatic updates is done in edgi-govdata-archiving/web-monitoring-ops#29.

PR #619, opened 2019-10-23
Calculate overall status code for a page

Mr0grog:
This will finish the work on #451.

We have a `status` field on pages, but we don’t actually set it right now. Because the occasional failure might happen when capturing a version, we don’t just take the latest version’s status. Instead, we only treat failures as real if they account for more than a certain percentage of the last N days.

Still to do:

- [x] Tests!
- [x] Clean up this logic a bit — there’s some duplication between handling candidate versions and the base version that we could get rid of.
- [x] Trigger this code at a relevant time (I’m thinking the end of an import job, similar to how we trigger automated analysis).
- [x] Create data migration to run this for all existing pages.


Mr0grog:
Update: I realized the `status` can change as time passes, even if no new versions were added, so this really has to be a job that runs on a schedule. My new approach is to make the `update_page_statuses` data migration a regular rake task, and give it a parameter to control how to behave. Then we’ll use a Kubernetes scheduled job to run that task. (Daily?)

Mr0grog:
This should pretty much be set to go.

PR #669, opened 2020-02-16
Add `content_length` & `media_type` to Version

### LINKED ISSUE #199: Add content length and content type to Version

Mr0grog:
This also includes a third field, `media_type_parameters`, which holds things like `charset=utf-8` that get tacked onto the end of the media type.

There is also a `content_type` getter/setter that includes the parameters, and sets `media_type` and `media_type_parameters` appropriately.

Fixes #199.

To do:

- [x] Add fields to Version model
- [x] Set `length` and `media_type` during import
- [x] Add data migration to update all existing versions

### DIFF
--- a/app/jobs/import_versions_job.rb
+++ b/app/jobs/import_versions_job.rb
@@ -100,7 +100,9 @@ class ImportVersionsJob < ApplicationJob
     elsif !Archiver.already_archived?(version.uri) || !version.version_hash
       result = Archiver.archive(version.uri, expected_hash: version.version_hash)
       version.version_hash = result[:hash]
+      version.length = result[:length]
       if result[:url] != version.uri
+        version.source_metadata = {} if version.source_metadata.nil?
         version.source_metadata['original_url'] = version.uri
         version.uri = result[:url]
       end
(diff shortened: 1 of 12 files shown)


Mr0grog:
The import job now sets `length` and `media_type` based on import record data. If not set, it will fall back to looking for `mime_type`, `encoding`, and `headers.Content-Length` in `source_metadta` for the relevant info (where we set that data in current imports).

Finally, if the archiver is able to load the version's body (either remotely or if it already has it in storage), it will override whatever is set for `length` and replace it with the actual length. This required adding `get_metadata()` to the file storage types. That's no major performance impact, though -- we already had to check whether the data existed in our storage.

Mr0grog:
This is probably ready to go. Going to take a break and come back with fresh eyes on it later. Otherwise, my biggest concerns here are mainly around whether the names are good. (e.g. `length` vs. `content_length`)

app/jobs/import_versions_job.rb
```diff
@@ -100,7 +100,9 @@ def import_record(record, row)
     elsif !Archiver.already_archived?(version.uri) || !version.version_hash
       result = Archiver.archive(version.uri, expected_hash: version.version_hash)
       version.version_hash = result[:hash]
+      version.length = result[:length]
```
Mr0grog: Yep, the more I think about it, the more sure I am this should be named `content_length` instead of `length`. The current name is too confusing since it often has a specialized meaning in many languages or tools.

Mr0grog:
Will merge and deploy this tomorrow morning, when the current import run is complete.

PR #671, opened 2020-02-19
Speed up Rubocop

bensheldon:
This should help speed up running the Rubocop linter.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
           command: bin/rails test:system test
       - run:
           name: Code linting
-          command: bundle exec rubocop
+          command: bundle exec rubocop --parallel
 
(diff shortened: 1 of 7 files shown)


Mr0grog:
Side note: not a big deal, but this is based on a way old commit! Trying to run this locally installed an old version of most of the dependencies.

bensheldon:
Thanks! I made a few more fix-ups. Let me know if there is anything else.

> This made me realize we have all the files in bin/ ignored, so Rubocop isn’t linting your changes here:

I removed the `bin/*` exclusion and fixed up the few lints based on what I thought was best (like ignoring an include). 

> Side note: not a big deal, but this is based on a way old commit! 

I'm embarrassed. I updated my origin and upstream and rebased off the latest code.

PR #672, opened 2020-02-19
Run AnalyzeChangeJob’s diffs in parallel

### LINKED ISSUE #420: AnalyzeChangeJob should run diffs in parallel
`AnalyzeChangeJob` currently requests and examines three separate diffs in series. Since they don’t depend on each other, we should see if our increased differ capacity lets us parallelize those requests.

Also see this guide on managing threading safely in Rails: https://guides.rubyonrails.org/threading_and_code_execution.html

bensheldon:
Uses Concurrent Ruby's Promises: https://github.com/ruby-concurrency/concurrent-ruby/blob/master/docs-source/promises.in.md

Closes #420

### DIFF
--- a/app/jobs/analyze_change_job.rb
+++ b/app/jobs/analyze_change_job.rb
@@ -81,9 +81,9 @@ class AnalyzeChangeJob < ApplicationJob
   # This may shortly become a simple call to some processing server endpoint:
   # https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/404
   def analyze_change(change)
-    text_diff_promise = Concurrent::Promise.execute { Differ.for_type!('html_text_dmp').diff(change)['diff'] }
-    source_diff_promise = Concurrent::Promise.execute { Differ.for_type!('html_source_dmp').diff(change)['diff'] }
-    links_diff_promise = Concurrent::Promise.execute { Differ.for_type!('links_json').diff(change)['diff'] }
+    text_diff_promise = schedule_diff('html_text_dmp', change)
+    source_diff_promise = schedule_diff('html_source_dmp', change)
+    links_diff_promise = schedule_diff('links_json', change)
 
     text_diff = text_diff_promise.value!
     source_diff = source_diff_promise.value!
@@ -264,5 +264,9 @@ class AnalyzeChangeJob < ApplicationJob
     end
 
     false
+  end
+
+  def schedule_diff(type, change)
+    Concurrent::Promises.future(type, change) { |t, c| Differ.for_type!(t).diff(c)['diff'] }
   end
 end
--- a/app/jobs/analyze_change_job.rb
+++ b/app/jobs/analyze_change_job.rb
@@ -81,39 +81,44 @@ class AnalyzeChangeJob < ApplicationJob
   # This may shortly become a simple call to some processing server endpoint:
   # https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/404
   def analyze_change(change)
+    text_diff_promise = Concurrent::Promise.execute { Differ.for_type!('html_text_dmp').diff(change)['diff'] }
+    source_diff_promise = Concurrent::Promise.execute { Differ.for_type!('html_source_dmp').diff(change)['diff'] }
+    links_diff_promise = Concurrent::Promise.execute { Differ.for_type!('links_json').diff(change)['diff'] }
+
+    text_diff = text_diff_promise.value!
+    source_diff = source_diff_promise.value!
+    links_diff = links_diff_promise.value!
+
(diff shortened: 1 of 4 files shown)


app/jobs/analyze_change_job.rb
```diff
@@ -73,39 +73,44 @@ def perform(to_version, from_version = nil, compare_earliest = true)
   # This may shortly become a simple call to some processing server endpoint:
   # https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/404
   def analyze_change(change)
+    text_diff_promise = Concurrent::Promise.execute { Differ.for_type!('html_text_dmp').diff(change)['diff'] }
+    source_diff_promise = Concurrent::Promise.execute { Differ.for_type!('html_source_dmp').diff(change)['diff'] }
+    links_diff_promise = Concurrent::Promise.execute { Differ.for_type!('links_json').diff(change)['diff'] }
```
Mr0grog: The docs say `Concurrent::Promise` has been deprecated and replaced with `Concurrent::Promises` (singular vs. plural for maximum hard-to-spot-ness). The new module has `#future` instead of `#execute`, but it looks like you’d otherwise use it the same here.

Since we create these and resolve them in the same scope and before touching `change` again, it’s probably not important, but maybe more safety-conscious to pass `change` as an argument.

Also maybe worth making a method for this like you did `diff_changes`?

```rb
schedule_diff(type, change)
  Concurrent::Promises.future(type, change) { |t, c| Differ.for_type!(t).diff(change)['diff'] }
end

**Comment by bensheldon** [comment]
@Mr0grog I added the simplest test case I could think of for `AnalyzeChangeJob`. Could you take a look at let me know what I've gotten wrong about it. My main goal intent was to test through the `Differ`, though I'm not quite sure if I got the setup correct. For example:

- I use the `changes` fixture. I wasn't sure whether I should use that fixture, or just make up a new pair of uuids because it appears that a new `Change` resource is created within this job.
- I hope my fake diff data is of the correct shape. I couldn't find an example of the expected data so I just molded something that fit the code.

File: test/jobs/analyze_change_job_test.rb [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+
+    differ_mock = Minitest::Mock.new(Differ)
+    def differ_mock.diff(_change)
+      {
+        'diff' => [
+          [1, 'addition'],
+          [-1, 'subtraction']
+        ]
+      }
+    end
```
Mr0grog: Yep, this is a good mock. There’s more stuff in the envelope object that wraps the diff, but it’s generally not relevant to this job. But if you’re curious, a more complete example would be:

```js
{
  "version": "1.2.3+hash",
  "type": "links",
  "change_count": 2,
  // other fields specific to selected diff type...

  // Diff is usually this format, but some diffs return something else, like
  // an HTML string (e.g. html_token). All the ones relevant for this test
  // follow this format, though. :)
  "diff": [
    [1, 'addition'],
    [-1, 'subtraction']
  ]
}
```

test/jobs/analyze_change_job_test.rb
```diff
@@ -0,0 +1,31 @@
+require 'test_helper'
+require 'minitest/mock'
+
+class AnalyzeChangeJobTest < ActiveJob::TestCase
+  test 'produces change with annotations' do
+    change = changes(:page1_change_1_2)
```
Mr0grog: > I use the `changes` fixture. I wasn't sure whether I should use that fixture, or just make up a new pair of uuids because it appears that a new `Change` resource is created within this job.

Ah, so this job is largely responsible for creating changes, so it might be better to create a new pair of UUIDs to analyze. (Technically the job makes an annotation, which makes a relevant change if one doesn’t already exist. But most of the time this job runs, that change doesn’t exist yet. A change mainly exists as a way to tie together and summarize multiple annotations.)

app/jobs/analyze_change_job.rb
bensheldon: Thanks for digging into the `Concurrent::Promise` docs and pointing me in the right direction. 👍

test/jobs/analyze_change_job_test.rb
bensheldon: Thanks! I updated the hash to match a little better in case I have this question again in the future.

test/jobs/analyze_change_job_test.rb
bensheldon: Done!

bensheldon:
@Mr0grog I addressed your requested changes:

- I added a test, and improved them to create a new set of versions to mock against
- I fixed my usage of `Concurrent::Promises` and added it to the Gemfile.

PR #673, opened 2020-02-21
Use `Promises.zip` to simplify promise code

Mr0grog:
Concurrent-ruby has a `zip` method to create one promise from several others -- its resolution value is an array of the promise results. Using it lets us remove a bunch of intermediary variables and simplify the code around promises as bit.

This is a follow-on to #672.


bensheldon:
🙌

PR #778, opened 2020-11-12
Remove `media_type_parameters`

### LINKED ISSUE #752: Refactor and normalize media type information
Right now, we track media type information on versions with two fields:
- `media_type`: The main type, e.g. `text/html` or `application/pdf`
- `media_type_parameters`: Any additional parameters for the type, e.g. `charset=utf-8; some-other=param`

I didn’t feel like this was ideal when I set it up, but was having trouble thinking of what would be better. In hindsight, I think we should ideally refactor this to:

- `media`: A normalized version of the main type. This is similar to the current `media_type`, BUT media types that are known to mean the same thing are normalized to a single name.

    For example, `application/html`, `application/xhtml`, and several others are all equivalent to `text/html`, the canonical type for HTML. What most applications really want to know in this case is “was it HTML?” and we should have a field that makes answering that easy. We have too much code all over the place that lists all the known “HTML” types and has to check against all of them. (There are other duplicate media types for non-HTML things; I’m less worried about handling those in the first attempt at this, although this should make those possible, too.)

- `charset` (or `encoding`?): The character encoding of the response body. This information is by far and away the most commonly used parameter, and has special importance, making it worth separating out. It also applies to a huge variety of media types, so it’s more cross-cutting than almost any other parameter.

- `media_type`: The full media type string, e.g. `text/html; charset=utf-8; some-other=param`. This is the true, canonical information that needs no interpretation or normalization (aside from the case normalization in #689).

    We might need to give this a different name so it doesn’t conflict with the existing `media_type` field. Maybe `media_type_full` or `content_type`?

This needs:
- A schema migration to create the new fields.
- A schema migration (later, in a separate PR) to remove the old fields.
- A data migration to fill in the new fields. (This will be a long, slow job, which is why it’s separate from the schema migration. See [`lib/tasks/data/`](https://github.com/edgi-govdata-archiving/web-monitoring-db/tree/main/lib/tasks/data).)
- Code in the `Version` model and `import_versions_job` to handle it.
- Updates to `analyze_change_job` to take advantage of it.

Mr0grog:
This removes the `Version#media_type_parameters` field (it wasn’t useful) and changes the `Versions#media_type` field to a cleaned-up, normalized, canonicalized field (instead of just reflecting whatever the HTTP response’s `Content-Type` header had). This makes it dramatically more useful without removing canonical information stored elsewhere.

Fixes #752.

### DIFF
--- a/app/assets/data/swagger.yaml
+++ b/app/assets/data/swagger.yaml
@@ -1384,9 +1384,6 @@ definitions:
       media_type:
         type: string
         description: The media type of the version, e.g. `text/html`.
-      media_type_parameters:
-        type: string
-        description: Parameters to the media type, e.g. `encoding=utf-8`.
       source_type:
         type: string
       source_metadata:
@@ -1917,9 +1914,6 @@ definitions:
       media_type:
         type: string
         description: The media type of the version, e.g. `text/html`.
-      media_type_parameters:
-        type: string
-        description: Parameters to the media type, e.g. `encoding=utf-8`.
       source_type:
         type: string
         description: An identifier for the source of this version data, e.g. `versionista`, `internet_archive`, etc.
--- a/app/jobs/analyze_change_job.rb
+++ b/app/jobs/analyze_change_job.rb
@@ -4,13 +4,8 @@ class AnalyzeChangeJob < ApplicationJob
   # text/* media types are allowed, so only non-text types need be explicitly
   # allowed and only text types need be explicitly disallowed.
   ALLOWED_MEDIA = [
-    # HTML should be text/html, but these are also common.
-    'appliction/html',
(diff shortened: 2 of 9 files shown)


Mr0grog:
Taking a break, but will give this a once-over with fresher eyes in a couple hours and merge, then deploy alongside #777.

Mr0grog:
This removes the `Version#media_type_parameters` field (it wasn’t useful) and changes the `Versions#media_type` field to a cleaned-up, normalized, canonicalized field (instead of just reflecting whatever the HTTP response’s `Content-Type` header had). This makes it dramatically more useful without removing canonical information stored elsewhere.

Fixes #752.

PR #793, opened 2020-12-05
Add `PageUrl` model to describe pages with multiple URLs

### LINKED ISSUE #492: Allow pages to have multiple URLs
**This is an idea/open question that I’d like to resolve relatively quickly.**

We’ve avoided figuring out how to best model pages that move or have multiple locations over time (and probably at the same time, too). While doing some routine updates today, I discovered that the domain http://www.cpc.noaa.gov/ was just shut down earlier this month. It looks like all those pages have been redirecting to identical paths at https://www.cpc.ncep.noaa.gov/ for quite a while (months, maybe longer?) and they’ve finally stopped responding to the older domain.

In the long term, we need to actually model this correctly. In the short term, we need to keep monitoring these pages, so my quick-and-dirty suggestion is to add a `notes` text field to the Page record in DB, update the `url` field for those pages, and add some human-readable text to the `notes` field describing the situation for for those pages.

Thoughts?

/cc @jsnshrmn @danielballan @gretchengehrke

Mr0grog:
This adds support for multiple URLs per page. The `url` field directly on a `Page` should now be treated as the current canonical URL, but pages may also have any number of additional URLs associated with them. Page + URL combinations may have a timeframe (`>= PageUrl.start_time` and `< PageUrl.end_time`) describing when they are valid and a notes field that are meant to mostly carry anecdotal information. The timeframe does have one technical purpose: if two pages occupied the same URL, we prefer the most current when adding new versions.

Things that still need to happen here:

- [x] Basic Migration
- [x] New class and associations
- [x] Tests for `PageUrl`
- [x] Data migration to populate new `PageUrl` records for existing pages
- [x] Tooling to merge pages
- [x] Update import logic to account for this (I *think* the only change we need is to add a `PageUrl` record if we found a non-exact page match [i.e. we matched against `url_key` and not `url`])
- [x] Add to the API (We may need to update the `?url=` query param on `PagesController`; not sure if we should return the URLs with page records or add a new controller for `/api/v0/pages/{id}/urls`).
- [x] Redirect merged pages to page they got merged into

Fixes #492.

### DIFF
--- a/app/controllers/api/v0/pages_controller.rb
+++ b/app/controllers/api/v0/pages_controller.rb
@@ -164,11 +164,12 @@ class Api::V0::PagesController < Api::V0::ApiController
 
     if params[:url]
       query = params[:url]
+      collection = collection.joins(:urls)
       if query.include? '*'
         query = query.gsub('%', '\%').gsub('_', '\_').tr('*', '%')
-        collection = collection.where('url LIKE ?', query)
+        collection = collection.where('page_urls.url LIKE ?', query)
       else
-        collection = collection.where(url: query)
+        collection = collection.where('page_urls.url = ?', query)
       end
     end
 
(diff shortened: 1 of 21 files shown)


Mr0grog:
This should be set to go!

PR #856, opened 2021-06-06
Refactor several poorly named fields on `Version`

### LINKED ISSUE #776: Refactor `Version` Model
The `Version` model has had a tortured history, starting as a local copy of Versionista’s data, and slowly evolving into something much different. Today it could probably use some cleanup.

Normally I’d say this is not worth the bugs it might introduce, but we developer involvement here is waning, and I think it’s valuable to do the work to make the data easier to learn and understand, whether that’s to make future development by other programmers easier to get started with or to make dead, archived data easier to use.

Current fields:

✅ = all good as-is, ⚠️ = should change, :x: = should remove, 🌱 = should add

| Status | Field | Category | Usage/Notes |
| ------ | ----- | -------- | ----------- |
| ✅ | `uuid`  | DB-centric | Acts as a convenient, single-field primary key, even though versions can also be uniquely identified by `(capture_time, capture_url, source_type)` |
| ✅ | `created_at` | DB-centric | Should never be modified; indicates when the record was created. |
| ✅ | `updated_at` | DB-centric | Indicates when the record was last modified. Only relevant insofar as we have non-canonical fields listed below. |
| ✅ | `capture_time` | Canonical | The time at which the version was originally captured in the source system. Component of the natural key. |
| ⚠️ | `capture_url` | Canonical | The URL this is a capture of, e.g. `https://epa.gov/climate-change/`. **I think we should rename this to `url`,** which would be more concise and clearer. Component of the natural key. |
| ✅ | `source_type` | Canonical | The system that originally captured this, e.g. `internet_archive`. Component of the natural key. |
| ✅ | `status` | Canonical | The status code of the HTTP response. |
| ⚠️ | `uri` | Canonical | The location of the captured HTTP response body. **I think we should rename this to `body_url`** for clarity. Everyone new gets tripped up on this one. |
| ⚠️ | `version_hash` | Canonical | Technically this is derived data, but it should be treated as canonical since we use it for fixity. **I think we should rename this to `body_hash`** for clarity, and to go with the suggested rename of `uri` to `body_url` above. |
| ✅ | `source_metadata` | Canonical | Arbitrary additional metadata from the source that we don’t have a good field for on `Version`. |
| ✅ | `page_uuid` | Derived | Note this is not canonical! Pages are really an arbitrary construct. We also want to start treating this as something user settable (rather than something exclusively calculated from underlying data), because pages having multiple URLs (#492) means pages are mergeable, and the page a version belongs to is a malleable concept. |
| ✅ | `title` | Derived | Title of the page/document/whatever, if we can figure it out. |
| ⚠️ | `content_length` | Derived | The actual size in bytes of the HTTP response body (not the `Content-Length` header). Doesn’t reside anywhere else in the DB, but is/can be automatically derived from the data found from the `uri` field. **I’d like to suggest renaming this to `body_length` or `body_size`** for consistency with the other `body_*` fields suggested above and to differentiate from the header of the same name, but don’t think it’s a big deal. The current name is not a problem. |
| ✅ | `media_type` | Derived | Currently the parsed media type from the `Content-Type` header. I don’t think the name should change, although we should consider sniffing content and canonicalizing the type here instead of just reflecting the header. (See also #752) |
| :x: | `media_type_parameters` | Denormalized | This is complicated to use, not typically of much value, and not worth surfacing directly on the model. If someone ever has an actual use for it, they should get it by consulting the headers. **We should remove it.** (See also #752) |
| ✅ | `different` | Denormalized | This is the only field that merely denormalizes data found elsewhere in the DB, but we currently use it in most of our queries. |
| 🌱 | `headers` | Canonical | **Does not exist/should add.** The headers of the HTTP response. Having this goes a long way towards clarifying `Version` as a canonical record of an archived HTTP response, which is what it should really be at this point. We do not always have headers (so this must be nullable, or maybe an empty object?), but when we do they should be here instead of in `source_metadata`. |
| ~🌱~ | ~`charset`~ | Derived | ~**Does not exist/should add.** Character encoding of the HTTP response body. **I’m less bullish on this one,** since we can set it in the response headers of `uri`/`body_url` and, for HTML, it is often specified (and specified more correctly) in the document itself. It can also be retrieved from the headers in cases where we didn’t have to sniff it during import.~ |

We also track redirects in `source_metadata` (so we know if the response is a direct response to `capture_url` or a response to a redirect from it), but I’m not sure that’s worth pulling up into the main model.

I’ve made a point of calling out DB-centric vs. derived vs. denormalized because I think these are particularly useful distinctions. If there weren’t practical concerns, I’d almost suggest the derived & denormalized data belongs in a second table that has a 1:1 relationship with versions. That way we are separating out the canonical record of an archived HTTP response from other useful data we want to make easily available about it.

If it weren’t a total mess to do, I’d also suggest renaming `Version` to `Capture` or `Snapshot`, but I don’t think that’s worth the complex mess and migration headaches. Renaming fields is small potatoes in comparison.

Mr0grog:
This is a first pass at the remaining items in #776. The `Version` model is pretty central to the whole system, so this is kind of a big and very tedious change. :\

This makes the following changes to `Version`:
- Rename `capture_url` → `url`
- Rename `uri` → `body_url`
- Rename `version_hash` → `body_hash`
- ~**Maybe** rename `content_length` → `body_length`~
- Add `headers` (a.k.a. move `source_metadata.headers` → `headers`)

Note that the `content_length` change hasn’t yet been implemented here. I’m a little on the fence. The current name is clear, and references a well known HTTP header. The proposed new name (`body_length`) is still clear and is more concise, but departs from well-known convention.

Remaining work to do here:

- [x] Migration
- [x] Docs
- [x] Main code changes
- [x] Add data migration to fill in `headers` from existing `source_metadata.headers` data
- [x] Add tests to ensure imports based on the old format don't break
- [x] Decide on `content_length` → `body_length`
- [x] Update seed files
- [x] Update downstream systems ([web-monitoring-processing][], [web-monitoring-task-sheets][], [web-monitoring-ui][], [web-monitoring-changed-terms-analysis][], [web-monitoring-versionista-scraper][] — not sure these last two really matter since they are not actively used anymore)

Fixes #776.

[web-monitoring-processing]: https://github.com/edgi-govdata-archiving/web-monitoring-processing
[web-monitoring-task-sheets]: https://github.com/edgi-govdata-archiving/web-monitoring-task-sheets
[web-monitoring-ui]: https://github.com/edgi-govdata-archiving/web-monitoring-ui
[web-monitoring-changed-terms-analysis]: https://github.com/edgi-govdata-archiving/web-monitoring-changed-terms-analysis
[web-monitoring-versionista-scraper]: https://github.com/edgi-govdata-archiving/web-monitoring-versionista-scraper


Mr0grog:
OK, I think I’m not going to change `content_length`. It’s pretty clear as-is, and the shortness/consistency of `body_length` is a really minor benefit vs. the issues with refactoring the API here.

Mr0grog:
Migrated the seed file with the following quick `jq` commands:

    $ cat db/seed_import.json | jq -c '. + {body_hash: .version_hash, body_url: .uri} | del(.version_hash) | del(.uri)' > db/seed_import.new.json
    $ mv db/seed_import.new.json db/seed_import.json

(No need to update `capture_url` because it’s not used here.)

Mr0grog:
Downstream update PRs:
- https://github.com/edgi-govdata-archiving/web-monitoring-ui/pull/749
- https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/727
- https://github.com/edgi-govdata-archiving/web-monitoring-task-sheets/pull/10

Not going to worry about -versionista-scraper and -changed-terms-analysis since they are no longer in active use.

Mr0grog:
OK, this should be good to go pending updates to downstream consumers.

Going to let this bake for a while and come back with fresh eyes and and a careful review tomorrow or later in the week.

danielballan:
Renames done in pairs like this that effectively change the meaning of term can create hard-to-find bugs:

* Rename capture_url → url
* Rename uri → body_url

Are we sure we like `capture_url` -> `url` better enough to risk it? Alternative would be deprecating `url` entirely so that anywhere it's left over we know unambiguously that it's wrong and what it should be changed to.

Mr0grog:
> Are we sure we like `capture_url` -> `url` better enough to risk it?

Ha! These are actually the ones I feel most confident about, even though you are right that they are the most technically risky. I feel confident about these for two reasons:
- Technical issues are unlikely, because `url` and `uri` are still different, so any existing code won’t suddenly be doing the wrong thing.
- This is resolving longstanding confusion and questions I’ve gotten based on the names; we’re aligning them with what people intuitively expect them to mean, which is the opposite of what they currently are:
    - `capture_url` is ambiguous about whether it represents the URL that was captured or the URL where the capture resides (another way to clarify this might be `capture_url` → `captured_url`). But I think people mostly expect `url` or `uri` to be the URL that was captured, which leads me to…
    - `uri` → `body_url` is similarly very clarifying because people often first expect that `uri` is representing the URL/URI *that was captured,* rather than *where the capture is stored.*

danielballan:
Those justifications are convincing. I had actually missed the `uri` / `url` distinction in my first read. Now it's clear that's a nonissue.

Mr0grog:
Going to merge this in an hour or so after weekly sheets are done building.

Mr0grog:
Deployed to staging and and ran the data migration there. Took 79 minutes, but otherwise worked great. Deploying to production now. :)

PR #946, opened 2022-03-14
Add API route for sampled list of versions

Mr0grog:
This adds a new API route at `/api/v0/pages/:page_id/versions/sampled` that gets a "sampled" list of versions — that is, one version per sampling period. Right now, the sampling period is hard-coded to one day, but we might change that in the future. The main idea here is to give people a list of versions to look at, but reduce the amount of data in a useful way for pages that we capture extremely often (e.g. `https://epa.gov`/ gets many captures every day). For most use cases, we don't want to look more granularly than each day.

Basically, instead of getting a list of versions, this gets a list of objects like:

```js
{
  time: "2022-01-01",
  version_count: 3,
  // A version object like we would get from /versions.
  // This is the latest `different` version in the sample period
  version: { ... }
}
```

The order of sample periods is always latest first.

This is part of https://github.com/edgi-govdata-archiving/web-monitoring/issues/161

### DIFF
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -14,6 +14,7 @@ class Api::V0::VersionsController < Api::V0::ApiController
   end
 
   def sampled
+    @sampling = true
     raise API::NotFoundError('You must provide a page to sample versions of.') unless page
 
     # TODO: support variable sample periods. Need to figure out a reference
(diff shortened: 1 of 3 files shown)


app/controllers/api/v0/versions_controller.rb
```diff
@@ -13,6 +13,49 @@ def index
...
+    raise API::NotFoundError('You must provide a page to sample versions of.') unless page
+
+    # TODO: support variable sample periods. Need to figure out a reference
+    # point for when those periods start.
+    # We don't support the complex filters and options #index does; we want to
+    # keep this as simple and fast as possible.
+    query = page.versions
+    # FIXME: this should probably have special pagination by date, since we
+    # don't want a sample group split across two responses.
+    paging = pagination(query)
```
Mr0grog, line 27: May not want to wait for this, since it will be messy, and for the endpoint to be useful we also need to call it from the -ui project. Probably not worth blocking that work on this.

PR #990, opened 2022-08-02
Upgrade from Raven to Sentry SDK

### LINKED ISSUE #837: Switch from Raven to sentry-ruby
Raven has long been deprecated; we should switch to sentry-ruby (and maybe also sentry-rails). Migration guide: https://docs.sentry.io/platforms/ruby/migration/

Mr0grog:
Sentry's "Raven" library has been deprecated and in maintentance mode for a long time. The newwer, full-featured, updated gem is "sentry_sdk". Fixes #837.


Mr0grog:
This also gets us tracing and performance info, which is super nice!

PR #1046, opened 2022-12-01
Clean up CircleCI

Mr0grog:
This updates from a deprecated Docker image for Postgres, but I also need to figure out what went wrong on the main branch here, since tests are failing in CI (maybe because of this? but not likely) but not locally. :(


Mr0grog:
Welp, this passed, so I guess the deprecation was the problem. Kind of surprising, though.

PR #1061, opened 2023-01-17
Only update the page title for non-error versions

### LINKED ISSUE #468: Add heuristics to identify error pages with 200 status codes
Some sites have error pages that don’t respond with actual error status codes (i.e. they have a `200` status code instead of `>= 400`). For example: https://www.eia.gov/tools/models/datatools.cfm

Since we are now elevating status codes to a first-class value, it might be nice to have some heuristics for identifying these sorts of pages and assigning more accurate statuses to them. (Or even just a way for analysts to inform us of the situation when they run across one.)

I *think* the best place for this work is here in DB as a job, but it’s possible we should be putting it in [-processing](https://github.com/edgi-govdata-archiving/web-monitoring-processing) and [-scraper](https://github.com/edgi-govdata-archiving/web-monitoring-versionista-scraper) instead.

I think there are probably a few other open questions here, too:
- Should this apply to a version’s status, or only a page’s?
- How should we store things when the heuristically assigned status doesn’t match the actual status? (I’m thinking we change the `status` field on a version to the heuristic one and add a new field to `source_metadata` so that it has both the actual response status and the heuristically assigned one.)
- And obviously we should come up with some good heuristics.

### LINKED ISSUE #751: Page titles should not be updated by versions that are errors
The `Version` model has an `after_create` callback that updates its page’s `title` field: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/213dc89341a59d975eae35e438a0c7953820b61b/app/models/version.rb#L107-L112

However, it shouldn’t update the page title if the version in question represents an error — the page model represents the conceptual page (not the URL) that we are tracking, and it’s unhelpful to rename the page to represent a temporary error, or worse, a “not found” message when the page is permanently removed. It makes it hard to tell what page we’re talking about.

For example, this page: https://monitoring.envirodatagov.org/page/9c85fc4f-c0ef-4327-ac4e-df2b6314917f/ba9b8e97-668d-4c92-bf95-d2ee5ffa46f2..0f4b06b2-6473-4ff6-a5e6-aada11f65cdd  
…shows the title “Page Not Found | USDA” when it should show “USDA | Office of the Chief Economist | Climate Change | Effects | Climate Change and Agriculture Report.”

Screen Shot 2020-09-08 at 10 28 46 AM

To do this, we should probably add a method on `Version` named `is_error?` or similar rather than directly checking the status code so that we can account for things like #468 in the future.

Mr0grog:
If a new version is an error (e.g. a 404 status code because the page was removed), don't update the page record's title with the title from the error response. This also uses the filename component of the URL as a fallback title in case no versions with a parsed title can be found.

Fixes #751.
Fixes #468.

To-Do:

- [x] Check for error statuses before updating
- [x] Fall back to URL-based title
- [x] Add data migration to reset all page titles

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -254,7 +254,7 @@ class Page < ApplicationRecord
     # Fall back to the filename from the page's URL.
     if new_title.blank?
       filename = /\/([^\/]+)\/?$/.match(url).try(:[], 1)
-      clean_name = CGI::unescape(filename || '')
+      clean_name = CGI.unescape(filename || '')
       self.update(title: clean_name)
     end
   end
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -241,6 +241,24 @@ class Page < ApplicationRecord
     end
   end
 
+  def update_page_title(from_time = nil)
+    candidates = versions.reorder(capture_time: :desc)
+    candidates = candidates.where('capture_time >= ?', from_time) if from_time
+
+    new_title = nil
+    candidates.each do |version|
+      new_title = version.sync_page_title
+      break if new_title
+    end
+
+    # Fall back to the filename from the page's URL.
+    if new_title.blank?
+      filename = /\/([^\/]+)\/?$/.match(url).try(:[], 1)
+      clean_name = CGI::unescape(filename || '')
+      self.update(title: clean_name)
+    end
+  end
+
   protected
 
   def news?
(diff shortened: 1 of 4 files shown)


Mr0grog:
Taking a slight pause here; I discovered that we wound up with some placeholder titles (e.g. "None", ""No title available") from Versionista in the distant past. Going to see about a data migration to remove those.

PR #1066, opened 2023-01-20
Use range-based pagination for versions

Mr0grog:
This is a *super rough* first cut at the logic for range-based pagination for versions (see #579). It’s ugly, problematic, and absolutely not mergeable as-is, but lets us test things out. Still needs:

- [x] Cleanup and encapsulation. (Maybe pull this stuff out into a concern so it can be used on other models.)
- [x] Migration to add the appropriate indexes (`(capture_time, uuid)`, `(created_at, uuid)` maybe also `(page_uuid, capture_time, uuid)`, `(page_uuid, created_at, uuid)`).
- [x] Tests.
- [ ] Maybe extend to pages? (Not sure it’s worthwhile, there aren’t critical problems there like there are for versions)

*Update 2023-01-25: I’m not going to extend this to other endpoints. While it would be nice, the real priority right now is doing the minimum required for https://github.com/edgi-govdata-archiving/web-monitoring/issues/168*

## Overall Summary

The slow performance of offset-based pagination (when you get a batch of results with a query like `SELECT xyz FROM table OFFSET = n LIMIT = m`) is a serious problem on the the `versions` table (really any large table, but in practice it’s just `versions` for this app). This solves the problem by using range-based pagination (skipping to the place you want in the result set using a `WHERE` clause that filters to a range of values from at the same field(s) you are sorting by).

For the versions table, that means doing queries based on a different minimum/maximum `(capture_time, uuid)` or `(created_at, uuid)` for each page/batch/chunk of results rather than using SQL's `OFFSET` clause to get to the chunk we want. This is more-or-less compatible with with any existing client, since the `next` link still works and the `chunk_size` parameter also works the same.

On the other hand, this introduces some new restrictions:

- You can only sort by one field, and it must be either `capture_time` or `created_at` instead of any field(s) you want. We need an index for the fields we want to sort by, and these two are the ones we chose.

- Clients shouldn’t usually be setting the `chunk` parameter on their own, and will run into problems if they do, since that field is no longer an integer (now it is either a version UUID or a combined `"<timestamp>,<uuid>"` string).

- The `?include_total=true` parameter is no longer supported for versions. There’s just no performant way to do it without additional caching of that field somewhere (`count()` in Postgres always requires a full table scan; the entire point of this change is to prevent us from doing large scans in order to paginate results).

There’s also plenty in here that’s not perfectly clean, and not totally generalized. If we wanted to utilize this kind of pagination for other models/controllers, it would need some work to make more abstract.

### DIFF
--- a/app/controllers/api/v0/versions_controller.rb
+++ b/app/controllers/api/v0/versions_controller.rb
@@ -6,40 +6,13 @@ class Api::V0::VersionsController < Api::V0::ApiController
 
   def index
     query = version_collection
-    # paging = pagination(query)
-    # versions = paging[:query]
-
-    links = {
-      first: paging_path_for_version(
-        params: request.query_parameters.except(:chunk)
-      ),
-      next: nil,
-    }
-
-    chunk_size = (params[:chunk_size] || PagingConcern::DEFAULT_PAGE_SIZE).to_i.clamp(1, PagingConcern::MAX_PAGE_SIZE)
-    query = query.limit(chunk_size)
-
-    start_point = params.dig(:chunk)&.split(',')
-    sort_key = sorting_params&.first&.keys&.first || :capture_time
-    sort_direction = sorting_params&.first&.values&.first || :asc
-    query = query.ordered(sort_key, point: start_point, direction: sort_direction)
-
-    if query.length == chunk_size
-      # There's no way to do "prev" without changing the sorting (could do, but
-      # meh) and no great way to get "last" here, either, so we just don't fill
-      # them in.
-      last_record = query.last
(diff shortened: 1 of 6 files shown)


Mr0grog:
⚠️ Some important, big deal changes here!
- This changes the default sorting for `/api/v0/pages/<uuid>/versions`! It is now `capture_time:asc` (or more importantly, the same as `/api/v0/versions`) instead of `capture_time:desc`.
- `?include_total=true` is no longer supported for versions. I can’t find anywhere we actually use this for versions (we definitely use it for pages) — probably because of the performance issue — so I think this is OK.

PR #1068, opened 2023-01-24
Add new indexes for range-based pagination

Mr0grog:
Before we can actually do range-based pagination in any usefully workable way, we need to have indexes in place that support it (partially because the primary key for versions is a UUID, which is unordered -- this wouldn't be such a big deal with bigint). These indexes will take a while to build, so it's good to get these in place ahead of time.

A future change will remove the old, non-compound indexes that will be redundant after these new ones finish building.

Addresses part of #571, and required before deploying #1066.

### DIFF
--- a/db/migrate/20230124182815_add_compound_time_uuid_indexes_on_versions.rb
+++ b/db/migrate/20230124182815_add_compound_time_uuid_indexes_on_versions.rb
@@ -1,0 +1,17 @@
+class AddCompoundTimeUuidIndexesOnVersions < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    # The versions table has gotten so big that offset-based queries are a real
+    # problem, so we need to paginate by the actual sort critera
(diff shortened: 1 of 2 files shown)


Mr0grog:
One thing to think about here: version queries set `different = true` by default. Experimentally, when the table is big enough, it will use the new indexes here over the existing partial index that is used to optimize `different` queries (not so when the table is small! Interesting.). That’s almost certainly the right thing, but worth considering whether it will deoptimize the default query! Some thoughts/options:

1. This is probably fine. I spot-checked a bit and it slows things down, but not supremely horribly.
2. We can stop doing `different` queries. One of the original use cases was in restricting imports, which we abandoned a long time ago. The other main use case was in keeping queries over long time periods light, which was ultimately better solved by date-based sampling (#946, #992).
3. Add additional partial versions of these compound indexes.

I’m thinking (1) is probably OK for now, and maybe I should also do (2), or at least switch the default from `true` to `false`. (3) is *probably* not worthwhile.

Issue #1070, opened 2023-01-26
Audit API for parameters that can heavily de-optimize requests


Mr0grog:
We want the API to be publicly readable, and there are a lot of parameters or options that can cause it to have unacceptable performance (e.g. causing full table scans on large tables, causing N+1 queries) if publicly accessible. We need to audit all the controllers and models and either:
- Remove non-essential ones
- Restrict access to essential ones

PR #1084, opened 2023-02-03
Block public access to some expensive parameters

Mr0grog:
Some parameters for the various `/versions` collections cause expensive queries, so they are disallowed for public, non-logged-in usage.

Solves part of #1070.

### DIFF
--- a/app/assets/data/swagger.yaml
+++ b/app/assets/data/swagger.yaml
@@ -229,7 +229,6 @@ paths:
           format: uuid4
         - $ref: '#/parameters/chunk'
         - $ref: '#/parameters/chunk_size'
-        - $ref: '#/parameters/include_total'
         - $ref: '#/parameters/sort'
         - name: capture_time
           in: query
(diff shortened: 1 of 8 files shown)


Mr0grog:
Thinking about this some more, it might still make sense to allow `include_change_from_*` for `#show`. We could do this by changing `block_params_for_public_users` to take a first argument indicating which methods:

```rb
# First argument indicates which methods:
block_params_for_public_users :all, [:bad, :params]
block_params_for_public_users [:index, :sampled], [:other, :bad, :params]

# Or as a hash:
block_params_for_public_users {
  all => [:bad, :params],
  [:index, :sampled] => [:other, :bad, :params]
}

# Or just provide a method to use with `before_action`:
before_action do
  block_params_for_public_users [:bad, :params]
end
before_action(only: [:index, :sampled]) do
  block_params_for_public_users [:other, :bad, :params]
end
```

Mr0grog:
This now covers all controllers. I also wound up going with a slightly different, more explicit style:

```rb
block_params_for_public_users actions: :all,
                              params: [:source_metadata, :status]
block_params_for_public_users actions: [:index, :sampled],
                              params: [
                                :include_change_from_previous,
                                :include_change_from_earliest
                              ]
```

PR #1088, opened 2023-02-09
Change int primary keys to bigint

### LINKED ISSUE #1067: Use bigint for all non-uuid primary keys
We have some models with UUID primary keys and some that use int. We should have used bigint for both, but that ship has long since sailed on the ones with UUID keys. We should still upgrade the int keys to bigint, though.

- [x] imports
- [x] invitations
- [x] users

Mr0grog:
Integer primary keys can cause problems over time on big tables, and in general it's just better practice to use bigints instead. Some of our larger tables use UUIDs instead of ints, so this only upgrades the tables that used ints. (Practically speaking, none of these tables are really *likely* to overflow their primary key sequence, but better safe than sorry here, especially with technical maintainers offboarding.)

Fixes #1067.

### DIFF
--- a/db/migrate/20230209200733_change_int_to_bigint_for_primary_keys.rb
+++ b/db/migrate/20230209200733_change_int_to_bigint_for_primary_keys.rb
@@ -1,0 +1,32 @@
+class ChangeIntToBigintForPrimaryKeys < ActiveRecord::Migration[7.0]
+  def change_serial_primary_key_type(table, key, key_type)
+    change_column table, key, key_type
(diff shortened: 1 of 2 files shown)


Mr0grog:
I noted `schema.rb` doesn’t record anything about the sequences, and since this doesn't make things `:bigserial`, was worried a fresh DB would potentially be missing the sequences.

Tried this and and it appears to create the sequences and hook them up to the PK columns correctly, though:

```sh
# NOTE: delete the old DB or change the DB name in database.yaml first!
rails db:create
rails db:schema:load
```

PR #1091, opened 2023-02-15
Add disabled notice to AnalyzeChangeJob

### LINKED ISSUE #1090: Disable auto-analysis of new versions
A big goal from long ago was to provide useful automated analyses of changes to help analysts and others prioritize what to look at. We originally did that with [`AnalyzeChangeJob`](https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/main/app/jobs/analyze_change_job.rb), which is scheduled automatically after every import and create annotation records for every consecutive pair of versions: https://github.com/edgi-govdata-archiving/web-monitoring-db/blob/f01ad684c5c4df2775d1ca12e3a37dd76e400794/app/jobs/import_versions_job.rb#L26-L36

However, the analysis provided by that job always fell short (for a variety of reasons), and the problems it was meant to solve were ultimately better handled by [web-monitoring-task-sheets](https://github.com/edgi-govdata-archiving/web-monitoring-task-sheets). `AnalyzeChangeJob` is still running after every import, though! It’s doing a lot of work and generating a lot of data nobody never uses, so it’s probably time to disable it.

(NB: in an ideal world, we’d have better integrated the task sheets stuff so it ran automatically and posted results as annotations available through the API here, but that never happened for a whole other set of reasons. At this point, we are [putting the project to rest](https://github.com/edgi-govdata-archiving/web-monitoring/issues/168), and it no longer makes sense to try and better integrate these tools.)

Mr0grog:
We've disabled `AnalyzeChangeJob` via configuration in production. Since we also expect to never make use of it again, I've added a note in the code (I opted not to delete it in case it needs to get turned on in the future -- someone could always go get it back from the git history, but since there is no technical maintainer to replace me, I'm worried that's too big of a barrier.)

Fixes #1090.

### DIFF
--- a/app/jobs/analyze_change_job.rb
+++ b/app/jobs/analyze_change_job.rb
@@ -1,3 +1,10 @@
+# NOTE: THIS JOB IS NOT IN ACTIVE USE.
+# The use cases it served were better solved by the task-sheets project
+# (https://github.com/edgi-govdata-archiving/web-monitoring-task-sheets).
+# That said, the outputs there were never integrated into the database here in
+# an automated way, which this job did.
+#
+# The job code here is preserved in case we need to turn it back on.
 class AnalyzeChangeJob < ApplicationJob
   queue_as :analysis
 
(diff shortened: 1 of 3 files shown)


Mr0grog:
For reference, the configuration change is in commit 6e61be326d91ff08edd22d8bf4cfc36a6844a5cd of the ops-internal repo (not stored on GitHub, since it contains secrets).

PR #1099, opened 2023-04-03
Upgrade Ruby to v3.2.2

Mr0grog:
This includes two security updates.


Mr0grog:
Looks like CircleCI doesn’t have an image for Ruby; this will have to wait a bit: https://github.com/CircleCI-Public/cimg-ruby/issues/121

PR #1104, opened 2023-08-10
Add support for archiving DB to SQLite

Mr0grog:
⚠️ **Work in progress!** ⚠️

This adds a rake command to export the contents of the DB into a SQLite file for public archiving. It's *mostly* a pretty straightforward copy of every table/row, but we skip tables that are irrelevant for a public data set (administrative things like GoodJob tables, users, imports, etc.), drop columns with user data, and do some basic conversions.

Part of edgi-govdata-archiving/web-monitoring#170

For changes/annotations, we probably want to just select relevant annotations, like the important changes (make sure we have them all in the DB first, see https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/main/web_monitoring/cli/annotations_import.py), and only import those and the changes they apply to.

### DIFF
--- a/app/lib/data_helpers.rb
+++ b/app/lib/data_helpers.rb
@@ -59,35 +59,6 @@ module DataHelpers
     total
   end
 
-  def self.iterate_batches(collection, by: nil, batch_size: 1000, &_block)
-    by ||= [collection&.primary_key&.to_sym || :uuid]
-    ordering = by.collect {|key| "#{key} ASC"}.join(', ')
-    order_where_placeholders = by.collect {'?'}.join(', ')
-    next_chunk_where = "(#{by.join(',')}) > (#{order_where_placeholders})"
-    total = 0
-    last_key = nil
-
-    loop do
-      chunk = collection.order(ordering).limit(batch_size)
-      if last_key
-        key_values = by.collect {|key| last_key[key]}
-        chunk = chunk.where(next_chunk_where, *key_values)
-      end
-
-      items = chunk.to_a
-      total += items.length
-      yield items unless items.empty?
-
-      break if items.length < batch_size
-
-      last_key = by.each_with_object({}) do |key, result|
-        result[key] = items.last.try(key)
-      end
-    end
-
-    total
-  end
-
   # Update many records with different values at once. (But it must update the
   # same *attributes* on each record.) This takes an ActiveRecord collection to
   # iterate over and gather the updates, then a list of the attributes that
--- a/app/lib/data_helpers.rb
+++ b/app/lib/data_helpers.rb
@@ -59,6 +59,35 @@ module DataHelpers
     total
   end
 
+  def self.iterate_batches(collection, by: nil, batch_size: 1000, &_block)
+    by ||= [collection&.primary_key&.to_sym || :uuid]
+    ordering = by.collect {|key| "#{key} ASC"}.join(', ')
+    order_where_placeholders = by.collect {'?'}.join(', ')
+    next_chunk_where = "(#{by.join(',')}) > (#{order_where_placeholders})"
+    total = 0
+    last_key = nil
+
+    loop do
+      chunk = collection.order(ordering).limit(batch_size)
+      if last_key
+        key_values = by.collect {|key| last_key[key]}
+        chunk = chunk.where(next_chunk_where, *key_values)
+      end
+
+      items = chunk.to_a
+      total += items.length
+      yield items unless items.empty?
+
+      break if items.length < batch_size
+
+      last_key = by.each_with_object({}) do |key, result|
+        result[key] = items.last.try(key)
+      end
+    end
+
+    total
+  end
+
   # Update many records with different values at once. (But it must update the
   # same *attributes* on each record.) This takes an ActiveRecord collection to
   # iterate over and gather the updates, then a list of the attributes that
--- a/lib/tasks/export_sqlite.rake
+++ b/lib/tasks/export_sqlite.rake
@@ -1,8 +1,24 @@
+# Convert Ruby types to SQLite. Used when writing Active Record models to SQLite.
 SQLITE_CONVERSIONS = {
   'TrueClass' => ->(_value) { 1 },
   'FalseClass' => ->(_value) { 0 },
   'ActiveSupport::TimeWithZone' => ->(value) { value.utc.iso8601(3) },
   'Hash' => lambda(&:to_json)
+}.freeze
+
+# Convert Postgres types' "text" representation to SQLite. This is used in COPY operations, while the above
+# SQLITE_CONVERSIONS is used in normal ActiveRecord-based operations.
+# If a type is not listed here, it will be passed to Postgres to parse as a string.
+PG_TEXT_SQLITE_CONVERSIONS = {
+  boolean: lambda { |value|
+    if value == 't'
+      1
+    elsif value == 'f'
+      0
+    end
+  },
+  # Truncate seconds to 3 decimal points.
+  timestamp: ->(value) { value&.sub(/(:\d\d\.\d\d\d)\d*/, '\1') }
 }.freeze
 
 def sqlite_convert(ruby_value)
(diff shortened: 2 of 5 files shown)


app/lib/data_helpers.rb
```diff
@@ -59,6 +59,35 @@ def self.iterate_time(collection, interval: nil, start_time: nil, end_time: nil,
...
+      break if items.length < batch_size
+
+      last_key = by.each_with_object({}) do |key, result|
+        result[key] = items.last.try(key)
+      end
+    end
+
+    total
+  end
+
```
Mr0grog, line 90: Now that we are on Rails 8, this is built into ActiveRecord. You can just do `<collection>.in_batches <block>` and specify the columns to use in the `cursor` argument. More info: https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches

lib/tasks/export_sqlite.rake
```diff
@@ -0,0 +1,275 @@
...
+        write_rows_sqlite(db, 'maintainerships', maintainership, [
+                            :maintainer_uuid,
+                            :page_uuid,
+                            :created_at
+                          ])
+      end
+    end
+
+    puts 'Writing versions...'
+    DataHelpers.iterate_batches(Version.all, by: [:capture_time, :uuid], batch_size: 10_000) do |versions|
```
Mr0grog, line 242: *Should* now be able to do:

```suggestion
    Version.in_batches(of: 10_000, cursor: [:capture_time, :uuid]) do |versions|
```

Or better yet, you get some extras for doing each record, like we do here:

```suggestion
    Version.in_batches(of: 10_000, cursor: [:capture_time, :uuid]).each_record do |version|
      write_rows_sqlite(...)
    end
```

Although this doesn’t let us wrap a SQLite transaction around each batch, so maybe not? Not sure the perf impact here.

Mr0grog (commented):
Need to come back around and actually get this done. Now that we’ve upgraded to Rails 8 there are some useful builtins and some code can be deleted.

Mr0grog:
Happy to discover I already imported all the sheet-based significant annotations to the DB after starting this PR (over a year ago, ugh): https://github.com/edgi-govdata-archiving/web-monitoring-processing/pull/853

So I just need to test this a bit more, but it should basically be ready to go.

Mr0grog:
The lesson of this weekend’s work trying to apply in production is:
- This approach works *great* for small databases, but not well for our production database (not huge, but still ~100 GB of data), where outputting the versions takes days, plural.
- In AWS, this is mainly constrained by disk IOPS, not memory or processing power (I figured more memory would at least help with indexing, but it doesn’t! The constraint there is still mainly read bandwidth).
    - Our production DB is on gp2 disks, which run out of burst capacity quick and then are *very* slow to read. Generating a dump of the DB, reloading it on a gp3 disk, and exporting from *that* DB instance is much faster, although still impractically slow (maybe more feasible if I maxed out the disk’s configurable IOPS? Not sure). It took 12 hours to get 20% of the way done, so not great.
    - **Side note:** we should probably upgrade the production DB to gp3 disks; it's just cheaper and more efficient at our DB size.
- Interestingly, `pg_dump` or `COPY` commands, that write out whole tables, still work really well under these constraints!
- Similarly, bumping up the number of rows returned in each batch from the `versions` table helps in this script, but I worry about hitting other problems as the batch size gets big.

There are interesting solutions like https://github.com/caiiiycuk/postgresql-to-sqlite, which reads a Postgres dump and reformats the data before importing to SQLite. It doesn’t work for us, though:
- It skips a bunch of data types we use. Most critically, it drops our JSONB columns. this is a deal-breaker.
- It creates tables with no column types. That’s OK since SQLite doesn’t really *enforce* column types, but it’s not great.
- It recreates a lot of indexes we don’t need in an archival copy (not that big a deal).

These would probably be solvable by modifying the source, but I’m having trouble getting the Scala toolchain installed right now (server issues with scala-sbt.org). And that seems overcomplicated, anyway?

Given the speediness of `pg_dump` and `COPY`, though, the process described by this repo (although the code is not *quite* what we want) seems to work alright: https://github.com/tantaman/pg2sqlite

However, the specific code there isn’t quite right on all counts. But following it’s approach:

1. Export tables as CSVs from Postgres. You can do this with the `COPY` SQL statement or the `\copy` instruction in psql.
    - `COPY` runs on the server, so you need to dump the DB and recreate it locally to be able to write a local file. The above linked repo does `COPY TO STDOUT` instead of to a file, and uses an efficient, streaming writer to write to the local disk. *Psycopg in Python has this built-in with the `copy_expert()` method; I’m not sure if ActiveRecord has anything similar. But we can just use Python for this if we want!* (Docs: https://www.postgresql.org/docs/current/sql-copy.html)
    - `\copy` in `psql` writes to local paths instead (it does the same trick as above, running `COPY TO STDOUT` and then streaming that output to disk in the client, but it’s built-in). (Docs: https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-META-COMMANDS-COPY)
    - **I skipped `changes` and `annotations`** and instead added them via the code from this PR, since we are only saving a small subset of these otherwise even-bigger-than-versions tables. But it looks you can also provide a complete query to the copy command, so you might be able to do that filtering the same way. *Also remember that we drop the `author_id` column from the annotations table, so you’ll need to do that here, too.*
    - So the command is something like `\copy pages TO 'csvs/pages.csv' WITH (FORMAT CSV, HEADER)`

2. Create a new sqlite DB file to import into.

3. Create the schema in sqlite using the same schema commands in the code in this PR (but make sure columns are in the same order as in the CSV exports!!! SQLite will not actually use the headers when importing, it will just import the columns in order).

4. Import the CSVs into the sqlite DB. use the aforelinked Python script or by hand in sqlite CLI: `.mode csv` then `.import --csv --skip 1 <filepath> <tablename>`. (`--skip` skips the header row; SQLite doesn’t know how to differentiate between it and the data, but it’s useful to have for cleaning and checking the results).

5. Run the code from this PR to export `changes` and `annotations`, since they do more limited queries.

6. Clean up the columns with incorrect types. SQLite’s import format doesn’t *quite* match Postgres’s output, and SQLite has dynamic data types — so if it can’t figure out how to make the import into the correct type for the column, it just stores it as a string. Not great. We have to clean up:
    - Nulls. CSV doesn't have anything for null, so we have to look for empty string values in non-string columns. On string columns, we’re just out of luck (but I don’t think there are any where this is that big a deal).
    - Booleans. Postgres outputs `t` and `f`, but SQLite wants `TRUE` or `1` and `FALSE` or `0`.
    - **NOT timestamps.** This is kind of a special case. We use the SQLite `DATETIME` column type, which is really an alias for `NUMERIC`. Buuuut SQLite handles ISO 8601 strings and numbers pretty transparently in any date functions. And its treatment of numeric datetimes is a little rough: in some cases, numbers are fuzzily handled as either julian days or unix epoch seconds depending on their value and in others numbers are treated as julian days by default (while most systems interpret as unix epoch) and its date math ignores leap seconds and other funky bits with time. For our *archival* purposes, storing the ISO 8601 timestamp is probably better — more clear, portable, and reliable. Postgres outputs ISO8601 timestamps for CSVs, which SQLite fails to turn into numbers and instead stores as strings, so these are already what we want them to be, even though it disagrees with the stated column type.

    - So it’s something like:

        ```sql
        UPDATE changes SET
            priority = CASE priority WHEN '' THEN NULL ELSE priority END,
            significance = CASE significance WHEN '' THEN NULL ELSE significance END;

        -- It might be nice to null these out since sqlite can't handle infinity for datetimes, but in practice it's fine,
        --   and these columns are currently declared non-nullable.
        -- UPDATE page_urls SET
        --     from_time = CASE from_time WHEN '' THEN NULL WHEN 'infinity' THEN NULL WHEN '-infinity' THEN NULL ELSE from_time END,
        --     to_time = CASE to_time WHEN '' THEN NULL WHEN 'infinity' THEN NULL WHEN '-infinity' THEN NULL ELSE to_time END;

        UPDATE pages SET
            active = CASE active WHEN 't' THEN TRUE WHEN 'f' THEN FALSE ELSE active END,
            status = CASE status WHEN '' THEN NULL ELSE status END;

        UPDATE versions SET
            source_metadata = CASE source_metadata WHEN '' THEN NULL ELSE source_metadata END,
            status = CASE status WHEN '' THEN NULL ELSE status END,
            content_length = CASE content_length WHEN '' THEN NULL ELSE content_length END,
            media_type = CASE media_type WHEN '' THEN NULL ELSE media_type END,
            headers = CASE headers WHEN '' THEN NULL ELSE headers END,
            different = CASE different WHEN 't' THEN TRUE WHEN 'f' THEN FALSE WHEN '' THEN NULL ELSE different END;
        ```

This is all stupidly complicated! I don’t like it. But other methods are so slow as to make meaningful archival impractical.

Mr0grog:
Some final bits here:

- Bumping the batch size up to 500,000 seemed to speed things up quite a bit. Much higher and it at all the RAM on the machine I was using (16 GB) and locked up. So for batched queries, really big values are optimal, but you have to be careful not to overflow available RAM.

- I figured out how to do the streaming `COPY` trick in Ruby, and that gets us something a *little* faster than really big batches as above and extremely low on memory cost. More complex code, but vastly improved overall.

- No matter what you really want to be reading from and writing to gp3 or fancier disks here. gp2 bogs down (unless it’s a huge drive!) quite fast when ripping through a whole table. However, the `COPY` method performs even better than the batched method on gp2 than on gp3. So if you must deal with gp2, you must must must use `COPY`.

Mr0grog:
This has since been done. I’m not sure I expect to repeat this process in the same way in the future, or even to do this at all (this was most meaningful for old data that was not originally Internet Archive-based, which this project should not have in the future), so I’m closing without merging.

PR #1110, opened 2024-04-23
Upgrade Ruby to v3.2.4

Mr0grog:
This resolves some security vulnerabilities (see https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-3-1-released/). Technically we should probably upgrade to v3.2.4 (same security patches, but for the 3.2 release line), but there's no obvious reason not to just move up to the 3.3 release line and no breaking changes.


Mr0grog:
CircleCI has not yet released an image for this version of Ruby. Tests pass locally, so I don’t expect to see any failures once CircleCI updates, but I’ll still wait for CI before merging just in case.

PR #1125, opened 2024-11-19
Upgrade Rails to v7.1

Mr0grog:
We may be spinning this app back up again for round two of Trump, so I’m starting to dust off the cobwebs by getting Rails more up-to-date. This will be the first of 3 (!) upgrades (then 7.2, then 8.0). Rails 7.0 is no longer actively maintained, so I am starting here, then will probably upgrade *other* deps, then go back to upgrading Rails.

This is a first pass; there’s still more to check over here:

- [x] Make sure autoloading is all good. We did a bunch of weird hijinx with autoloading at various points in time, and Rails 7.1 made some other changes. I think we are fine as long as we don’t call `autoload_lib`, but this needs more testing. (Ultimately we should clean up our autoloading story; I understand better now the ways one “should” use `lib` and `app/lib`, although half the problem is that Rails core folks don’t even have their own entirely clear/agreed-upon story there.)
- [x] Update all the defaults and remove `config/initializers/new_framework_defaults_7_1.rb`.
- [x] Make sure I’ve got the settings right w/r/t the new `assume_ssl` config var.
- [ ] Maybe remove Spring? It’s not so common these days, and I noticed that the default `test` config now does things that breaks it.
    - *Update: I think removing Spring is a good idea (fewer moving parts, it has a lot of edge cases, and this version of Rails included new defaults that break it and it references settings that are deprecated, which seems to indicate that it’s on the way out as a common tool). HOWEVER, I think it’s better to drop it in a separate PR/commit.*
- [x] Check over new features we should be using?
    - Active Record `normalizes` is nice, and IIRC there are some spots where we should probably be using it.


app/controllers/api/v0/maintainers_controller.rb
```diff
@@ -47,7 +47,7 @@ def create
           conditions = data.select {|key, _| valid_fields.include?(key)}
           begin
             maintainer = Maintainer.find_or_create_by(conditions)
-          rescue ActiveRecord::RecordNotUnique
+          rescue ActiveRecord::RecordNotFound
```
Mr0grog, line 50: The behavior of `find_or_create_by` changed in 7.1, and it’s now more like `find_or_create_or_find_by`. (https://github.com/rails/rails/issues/51149) Kinda weird. Maybe want to modify how we do this operation to work differently now for efficiency’s sake.

Maybe just:

```rb
maintainer = Maintainer.find_by(conditions) || Maintainer.create!(conditions)
```

Or be more complicated, and find by name, then if `parent_uuid` is set, assert that it matches. 🤷 

Either way, Rails core folks don’t view our usage here as a pattern they intend to support (it *is* probably unusual).

Mr0grog:
Final notes:
- Going to remove Spring in a separate, follow-on PR.
- The one place where `normalizes` sort of made sense had some additional complexity with updating *other* derived properties, and didn’t seem worth refactoring.

PR #1155, opened 2024-11-27
Upgrade Rack to v3 + Rack-related gems

Mr0grog:
There are a bunch of intertwingled Rack-related gems that require either Rack < 3 or >= 3, so they have to be updated in a coordinated way. Prerequisite for #1146.


Mr0grog:
Nevermind, I just worked my way to an identical change with #1146.

PR #1175, opened 2025-01-03
Upgrade Ruby to v3.4.1

Mr0grog:
Upgrade to the latest version of Ruby.


Mr0grog:
Well, looks like this is failing because CircleCI has not published an image for Ruby 3.4.0 or 3.4.1 yet, even though it’s been out for over a week. Will have to wait on that. :\

PR #1176, opened 2025-01-10
Fix e-mail invitation expiration and confirmation issues

Mr0grog:
This fixes a few issues with user invitations/registration. The expiration date on an invitation somehow wound up being calculated based on N *seconds* instead of *days* from the current time!

~This also disables user e-mail confirmation for now. We’ve got some issues with Postmark that I *think* are on their end (I have a support ticket open with them), so it’s better not to require any e-mail to get an account set up for now. (Confirmation was always a weird step since invitations are currently required here; it was originally intended for a future vision where we wouldn’t require invitations, but that future will probably never exist for this codebase.)~

### DIFF
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -68,7 +68,7 @@ class Invitation < ApplicationRecord
 
   def ensure_expiration
     unless self.expires_on
-      self.expires_on = Time.now + EXPIRATION_DAYS
+      self.expires_on = Time.now + EXPIRATION_DAYS.days
     end
   end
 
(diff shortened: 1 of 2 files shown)


Mr0grog:
Obviously there is some more complex cleanup to do here. It’s too late to finish this tonight in a sane mindset, and this is not actually an emergency.

I should probably:
1. Separate out the basic stuff from the first two commits and push that.
2. Give this confirmation-disabling setup a more thorough consideration. Maybe the Postmark issue will get fixed for me tomorrow and this won't matter anyway. 🤷

Issue #1208, opened 2025-02-28
Remove default `different` check from API


Mr0grog:
When looking at `versions`, we only return records where the `different` column is `true` by default. Setting `?different=false` in the API disables this check, and does not look for versions where `different` is `false`. This is confusing, but more importantly, it’s based on a bad assumption we made early on: that most of the time, you’d only want a list of versions that were really “versions,” where something had changed. We realized really quickly this was wrong, and almost all our tools and scripts have to go out of their way to set `different=false`. When writing new tools, this often catches people (including me, who created this situation!) unaware and causes problems when they forget to set it to `false`. Fixing this is long overdue.

We should:
1. Just return all versions if `?different` is not set in a request.
 
2. Consider removing this field from the database entirely. It imposes a lot of extra work on imports that come out of order, which is extremely common now! We originally thought this would be rare and unusual.

    The theory behind this field was making it easy to know when there was a new “version,” or rather, to know when the page changed. This field doesn’t actually do that very well:
    1. Knowing how closely the timestamp of the “different” version matches the *actual* time of a change requires knowing when the previous version was, the previous version may not have been different from *its* preceding version, so it could be hidden in a normal query. If we really want a shortcut for this, this field should probably be the time of the preceding version instead of a boolean.
    2. In actual reality, a lot of pages will have some unique session ID or other random or frequently changing value in the HTML that has no impact on the page content, but causes it to be considered “different” on every check (since “different” is based on the hash of the response body). That makes this field largely pointless on a lot of pages. Mostly we just see it working “correctly” for PDF files.

(2) is more speculative, and should probably be in a separate PR and release than (1), which is more straightforward.

Mr0grog:
Did a quick search and it looks like almost all our projects that interface with the API always set `different=false`. The exception is `versionista-scraper`, which is not really actively maintained, and hails from a time when we hadn’t yet realized how poorly this works.

Seems like good evidence this behavior in the API should be turned off.

Mr0grog:
I more-or-less took care of step 1 in #1232. However, I did retain usage of the `different` field (*not* the query param — those are all gone) in one place: when getting `/pages?capture_time=<from_time>...<to_time>`, it returns pages that have versions with `different=true` that were captured in that time range. This was the default before (you could disable with `?different=false`) and is now the only way it works.

[...]

PR #1232, opened 2025-04-04
Drop support for `?different` query param

Mr0grog:
This covers part 1 of #1208 and generally removes support for the `different` query param, which doesn't behave like you would probably expect and regularly causes problems. It was designed in a world where we expected to have such frequent and continuous captures of pages that any time we had a "different" version, we knew it had *just* changed. The reality is that you can never get captures frequent enough to do that even in practical terms, and in theoretical terms it's just wrong. We've long since switched to always querying in a way that removes checking the `different` field, and it's past time to remove it from the API.

This leaves it in place in ONE spot of the API, which does not return an actual list of versions and thus does not have the same problems: `GET /pages?capture_time=abc...def` implicitly checks the `different` column because the goal is to tell you what pages had observed changes in that time period. We could back off that idea and just have it check whether there were any captures, but I think the valuable use cases for that are pretty rare. But maybe I’ll wind up revisiting that.

This also drops `Version.invalid_changes`, which I noticed is:
1. Not (no longer?) definitionally sensible, and
2. Not used anywhere.

### DIFF
--- a/app/assets/data/swagger.yaml
+++ b/app/assets/data/swagger.yaml
@@ -130,7 +130,9 @@ paths:
           type: string
           in: query
           description: >
-            Only retrieve pages with versions captured during this time frame. Should be in the format `{start date}..{end date}`, where dates are ISO 8601 formatted. Either date can be left out to make the query open-ended.
+            Only retrieve pages with observed differences (versions that differed in some way from the previous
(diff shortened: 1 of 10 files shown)


Mr0grog (commented):
Still looks good with fresh eyes today, and the messy migration of version headers is done now. Should be good to merge!

PR #1268, opened 2025-07-17
Upgrade Ruby to v3.4.5

Mr0grog:


Mr0grog:
Available on Circle now and passes!

Issue #1351, opened 2026-02-23
Add ability to “merge” tags


Mr0grog:
Because tags are a bit messy by nature, we occasionally need to clean some up by merging synonymous tags together (for example, this morning I merged `agency:FWS`, `agency:USFWS`, and `agency:U.S. Fish and Wildlife Service`). This is all stuff that can be scripted through the API, but it would be nice to have a more straightforward shortcut here. We also eventually want to build some self-service admin tooling for folks to add/remove pages, tags, and maintainers, and this would support that, too (see “Non-Critical Goals” from the [2025 Q4 roadmap](https://github.com/edgi-govdata-archiving/web-monitoring/issues/184)).

For now, this doesn’t need to be exposed in the API; it’s just for jobs and/or console use.

My quick-and-dirty ruby script for this is something like:

```rb
def merge_tags(final_tag, old_tag)
    Page.joins(:tags).where(tags: {name: old_tag}).each do |p|
        p.add_tag(final_tag)
        p.untag(old_tag)
    end
    raise RuntimeError("Failed to remove tag '#{old_tag}' from all pages") if Page.joins(:tags).where(tags: {name: old_tag}).count > 0
    Tag.where(name: old_tag).first.destroy
end
```

That would probably become a class method on `Tag`, though it needs to be more generic (taggability is polymorphic, so it shouldn't reference `Page`), and could probably also be more generic.

PR #1385, opened 2026-04-09
merge tag script

marsherrer:


marsherrer:
addresses https://github.com/edgi-govdata-archiving/web-monitoring-db/issues/1351

app/models/tag.rb
```diff
@@ -5,4 +5,22 @@ class Tag < ApplicationRecord
 
   has_many :taggings, foreign_key: :tag_uuid
   has_many :pages, through: :taggings, source: :taggable, source_type: :Page
+
+  def self.merge(final_tag, old_tag)
+    final_tag = find_or_create_by(name: final_tag.strip) unless final_tag.is_a?(Tag)
+    old_tag = find_by(name: old_tag.strip) unless old_tag.is_a?(Tag)
+
+    return if old_tag.nil?
+    return if old_tag == final_tag
```
Mr0grog, line 14: I think it would be better to raise an exception in these cases.

app/models/tag.rb
```diff
@@ -5,4 +5,22 @@ class Tag < ApplicationRecord
...
+
+    old_tag.taggings.each do |tagging|
+      if Tagging.exists?(taggable_uuid: tagging.taggable_uuid, tag_uuid: final_tag.uuid)
+        tagging.destroy
+      else
+        tagging.update(tag_uuid: final_tag.uuid)
+      end
+    end
+
+    old_tag.destroy
```
Mr0grog, line 24: I think this could probably be done _much_ more efficiently I just 3 queries by:

1. Use a SQL subquery to upsert (i.e.`INSERT... ON CONFLICT DO NOTHING`) new `Taggable` records for the final tag that match all the attributes for the old tag's taggables.
2. Delete all the old tag's taggables.
3. Delete the tag (possibly you could do this instead of step two if we have cascading/dependent deletes, but I don't remember how we have these set up right now).

Also, this should probably be wrapped in a transaction or a lock, since these are potentially big queries that could wind up in a race condition with other updates to tags.

Mr0grog (changes requested):
Hey @marsherrer, thanks for taking this on! 🙇 

Sorry for my inattentiveness and slow reply — I'd thought it wasn't ready for review since you marked it as a draft and then I guess I lost track of it.

Could you please add some tests? I've also noted a few more things inline.

PR #1411, opened 2026-06-11
Upgrade Ruby to v4.0.5

Mr0grog:
Moving on up to Ruby 4.


Mr0grog:
This also drops Pry. I added originally for the nicer editing experience (syntax highlighting, multiline, completions), and regular old irb has since added most of those features. I don’t make much use of the other fancier parts of Pry, so it’s probably not really worth keeping around, waiting to cause problems like it did here.

PR #1426, opened 2026-06-23
Set effective status for ASP.net error redirects

Mr0grog:
Old-school ASP.net error handling still allows you to configure the server to redirect to a special error page, which often will send a 200 status code instead of something more appropriate.

By default (not even sure if you can customize it) it will include the path where the error was originally encountered in a query param named `aspxerrorpath`. That doesn’t really tell us what kind of error, but at least we know it shouldn’t be 200!


Mr0grog:
Useful example case…
- In scanner: https://monitoring.envirodatagov.org/version/61e77b72-adb6-446f-b335-10794e0be118
- In API: https://api.monitoring.envirodatagov.org/api/v0/versions/61e77b72-adb6-446f-b335-10794e0be118

PR #1443, opened 2026-07-25
Upgrade Bundler to v4.0.17

### LINKED ISSUE #1441: Upgrade Bundler to v4
Main announcement and breaking changes: https://bundler.io/v4.0/whats_new.html

This should hopefully be pretty smooth for us; I don’t think we are using any features that have been fundamentally changed or removed.

Mr0grog:
Bundler v4 overview: https://bundler.io/v4.0/whats_new.html

I still need to do some reading and make sure we don’t need to change any commands we use in the Dockerfile or README. Will probably hold off on merging until at least Sunday when I have more time for it and to address any failures that might happen when deploying.

Fixes #1441.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ WORKDIR /app
 # will be cached unless changes to one of those two files
 # are made.
 COPY Gemfile Gemfile.lock .ruby-version ./
-RUN gem install bundler && bundle install --jobs 20 --retry 5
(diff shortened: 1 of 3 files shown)


Mr0grog:
As far as I can tell, everything else we use is stable and this is good to go.

Mr0grog:
Nevermind, a few more tweaks. 🙃

